PR #30, opened 2015-01-03
clearer definition of transaction

CJentzsch:
To me, an actor external to Ethereum is not a very clear definition. A transaction must come from a node in the network. I think this is a clearer definition, since an external actor can be almost everything. Also the examples of an external actor (mobile, desktop, server) is only a subset, any kind of computer connected to the network (="a node in the Ethereum network") can send a transaction (IoT devices, ...).

### DIFF
--- a/Paper.tex
+++ b/Paper.tex
@@ -229,7 +229,7 @@ where $v$ is the account validity function:
 
 \subsection{The Transaction} \label{ch:transaction}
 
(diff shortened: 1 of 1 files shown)


gavofyork:
Kind of misses the point - the distinction is that they are pieces of formal instructions that are created and exist outside the scope of Ethereum. Altered the text to take on-board your issue, though.

CJentzsch:
I see, looks good now.

PR #45, opened 2015-01-20
changed block time from 8 to 12 seconds

CJentzsch:
unless some network propagation time should be taken into account, I think we should put a 12 here if you want a 12 second block time.


gavofyork:
actually no - being an exponential distribution, the median is about 70% of the mean. since the mean we're targetting is 12s, the median to target (which is what counts here) needs to be around 70% of that which is 8.5s.

PR #50, opened 2015-01-22
double mentioning of to-address in post provisional creation state

CJentzsch:


gavofyork:
This line refers to the \Theta, the message-call transition function. They're there for a reason - first is the receiving address, second is the code address. In this case, the two are the same.

PR #67, opened 2015-02-18
Add block number validation to block header validity function V(H)

Gustav-Simonsson:


CJentzsch:
Duplication of https://github.com/ethereum/yellowpaper/pull/64

PR #68, opened 2015-02-21
Add call depth limit condition to exception handling function

Gustav-Simonsson:


Gustav-Simonsson:
As discussed with @gavofyork this condition is already present in the definitions of CREATE, CALL and CALLCODE instructions in Appendix H and is of a different failure mode compared to the conditions in the exception handling function.

PR #72, opened 2015-02-21
Add missing wedges, ifs and use if instead of else/when

Gustav-Simonsson:


Paper.tex
```diff
@@ -1228,7 +1228,7 @@ \section{Recursive Length Prefix}\label{app:rlp}
 \begin{eqnarray}
 R_b(\mathbf{x}) & \equiv & \begin{cases}
 \mathbf{x} & \text{if} \quad \lVert \mathbf{x} \rVert = 1 \wedge \mathbf{x}[0] < 128 \\
-(128 + \lVert \mathbf{x} \rVert) \cdot \mathbf{x} & \text{else if} \quad \lVert \mathbf{x} \rVert < 56 \\
```
gavofyork: else if important here since it implies ordering of conditionals.

Gustav-Simonsson:
@gavofyork Added back the required "else if"

PR #74, opened 2015-02-24
message call to account with no code

CJentzsch:
May be I misunderstand something here. But why was the previously defined criteria for doing nothing but transfering value at a message call, that the output data is empty?
I can make a transaction to a contract with code, which needs to get executed, also without having any output data. Therefore I changed it to "account with no code".


CJentzsch:
I misunderstood something. o is not exactly the same as H_return ...

PR #86, opened 2015-03-17
Correct CODECOPY and EXTCODECOPY zero mem write

Gustav-Simonsson:


Paper.tex
```diff
@@ -1708,7 +1708,7 @@ \subsection{Instruction Set}
 \midrule
 0x39 & {\small CODECOPY} & 3 & 0 & Copy code running in current environment to memory. \\
 &&&& $\forall_{i \in \{ 0 \dots \boldsymbol{\mu}_\mathbf{s}[2] - 1\} } \boldsymbol{\mu}'_\mathbf{m}[\boldsymbol{\mu}_\mathbf{s}[0] + i ] \equiv
-\begin{cases} I_\mathbf{b}[\boldsymbol{\mu}_\mathbf{s}[1] + i] & \text{if} \quad i < \lVert I_\mathbf{b} \rVert \\ \text{\small STOP} & \text{otherwise} \end{cases}$\\
```
gavofyork: `STOP` and `0` have previously been declared equivalent.

Paper.tex
```diff
@@ -1719,7 +1719,7 @@ \subsection{Instruction Set}
 \midrule
 0x3c & {\small EXTCODECOPY} & 4 & 0 & Copy an account's code to memory. \\
 &&&& $\forall_{i \in \{ 0 \dots \boldsymbol{\mu}_\mathbf{s}[3] - 1\} } \boldsymbol{\mu}'_\mathbf{m}[\boldsymbol{\mu}_\mathbf{s}[1] + i ] \equiv
-\begin{cases} \mathbf{c}[\boldsymbol{\mu}_\mathbf{s}[2] + i] & \text{if} \quad i < \lVert \mathbf{c} \rVert \\ \text{\small STOP} & \text{otherwise} \end{cases}$\\
```
gavofyork: `STOP` and `0` have previously been declared equivalent.

Gustav-Simonsson:
OK, would it not still make sense to define the same in all three cases, whether it's STOP or 0?

gavofyork:
the other cases being?

Gustav-Simonsson:
@gavofyork CODECOPY and EXTCODECOPY currently have STOP whereas CALLDATACOPY have 0. Would be nice to align these three instructions to have all use either STOP or 0.

gavofyork:
CALLDATACOPY is not dealing with opcodes, hence 0. the other two exclusively deal with opcodes, hence STOP.

Gustav-Simonsson:
@gavofyork You previously stated that STOP and 0 are declared equivalent in the paper. If they are declared equivalent, why is there a difference in terms of whether the instruction is dealing exclusively with opcodes or not?

Also, CALLDATACOPY can deal with opcodes. A contract writer may send op codes in the input data to a contract and then use CALLDATACOPY to load it to memory and then use those opcodes as the code for a new contract with CREATE.

gavofyork:
It's not - it's designed to ensure consistency within domains. CALLDATACOPY *may* deal with opcodes, but in general it's just unrestricted byte data. Similarly CODECOPY *may* deal with unformatted data, but in general will be opcodes. When dealing with elements of CODE, we naturally use opcode literals notation, since they're useful for reasoning about semantics. When dealing with elements of bytes arrays we typically fallback to numeral literals for their generality.

sabr-p:
Hi, newb here :)
@gavofyork , you say:
 "CALLDATACOPY is **not** dealing with opcodes, hence 0. the other two **exclusively** deal with opcode ..." up above, and:
" CALLDATACOPY **may** deal with opcodes, but in general it's just unrestricted byte data. Similarly CODECOPY **may** deal with unformatted data,... "  in next comment.

insight/hindsight is 20/20, but anyone who relies on the docs (say, a random newb ;) ) will get confused with these conflicts. also, with the "mays", i read it it as CALLDATACOPY may(or not) use opcodes. CODECOPY may(or not) use data. which makes them basically equivalent, and that's not the plan, right?

gavofyork:
...and thus we come to the imprecisions and flaws of the English language. Within acceptable bounds of semantic precision, all the above statements are true:

Basically, when we interpret the contents of the `code` series, we only ever interpret it in terms of opcodes. Though it may, reasonably, include non-opcode data doesn't change the fact that within the confines of the YP it is only interpreted as a series of operations.

Similarly, the data is only ever interpreted, within the confines of the YP, as unformatted byte data; it does not deal with opcodes per se. Rather, someone could attribute that meaning *at a higher level* to them (which the YP does not and cannot know about).

It's all about levels of meaning: at the YP's level of semantics, the stricter statements are true. When viewed from other, higher levels, we happen to know that it's not quite as clean a distinction, but still, that's irrelevant from the point of view of the YP.

The YP does not try to present (or imply through the hinting of its literals) all levels of this system. Rather it attempts to extol only as much information as is required to understand the core consensus mechanism.

Gustav-Simonsson:
@gavofyork cool, thanks for the explanation! closing PR.

PR #96, opened 2015-04-01
Rename uncle to aunt to reduce androcentrism

Gustav-Simonsson:


obscuren:
:+1:

LefterisJP:
nice one :)

thehighfiveghost:
:P

alexvandesande:
The correct gender neutral term for Aunt and Uncle is "Ommer" according to [non-binary](http://nonbinary.org/wiki/Gender_neutral_language#Family_Terms).

debris:
made my day :+1:

chriseth:
I like this change. Such an innovative text should not be backward when it comes to encouraging diversity. Another solution would be to step away from the family tree analogy altogether and use a term like "branch" or similar.

PR #102, opened 2015-04-07
Adjust symbol used for ommer hash to name change.

chriseth:


gavofyork:
this then does read "ho", but never mind.

PR #106, opened 2015-04-08
max call stack (1024) does not result in an exceptional halting

CJentzsch:
,but instead just returns 0, as handeled in the CALL/CREATE instructions


CJentzsch:
Included in https://github.com/ethereum/yellowpaper/pull/107

PR #108, opened 2015-04-08
bloom size 64 -> 256 bytes

CJentzsch:
I just repalced 64 bytes with 256 bytes and 512 bits with 2048 bits. Is this enough, or has the bloom filter function changed?


winsvega:
Such change would require to review and recreate all bloom tests.
(like in invalidTransaction/Block  RLP/Headers)
remind me when such changes would be done to the cpp

CJentzsch:
its already implemented. The YP was outdated on this issue.

PR #112, opened 2015-04-09
Correct ranges for ECDSA privkey and signature r, s values

Gustav-Simonsson:
allowed privkey range is defined in Appendix A.1.1. in Suite B Implementer’s Guide to FIPS 186-3 (ECDSA) https://www.nsa.gov/ia/_files/ecdsa.pdf

allowed values for signature r and s values is defined in same doc in section 3.4.2
ECDSA Signature Verification

### DIFF
--- a/Paper.tex
+++ b/Paper.tex
@@ -1416,7 +1416,10 @@ g_r &=& 1 + \Big\lceil \dfrac{|I_\mathbf{d}|}{32} \Big\rceil\\
 
 The method of signing transactions is similar to the `Electrum style signatures'; it utilises the SECP-256k1 curve as described by \cite{gura2004comparing}.
 
-It is assumed that the sender has a valid private key $p_r$, a randomly selected positive integer in the range $(0, 2^{256})$ represented as a byte array of length 32 in big-endian form.
+It is assumed that the sender has a valid private key $p_r$, a randomly selected positive integer
+% to avoid line break in range definition
+
+in the range $(1, \mathtt{\tiny secp256k1n} - 1)$ represented as a byte array of length 32 in big-endian form.
 
 We assert the functions $\mathtt{\small ECDSASIGN}$, $\mathtt{\small ECDSARESTORE}$ and $\mathtt{\small ECDSAPUBKEY}$. These are formally defined in the literature.
 \begin{eqnarray}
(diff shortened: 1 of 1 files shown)


winsvega:
It is true. But may be there was a reason s being defined as     s < secp256k1p  ?

subtly:
And for V: https://github.com/ethereum/pyethereum/blob/master/pyethereum/transactions.py#L29

CJentzsch:
@vbuterin do you agree on this PR.

subtly:
@Gustav-Simonsson Might also look at updating ranges for v in order to support recovering deterministic signatures. (see python link)

v indicates if nonce's r was greater than order of the curve prior to mod/multiply w/the message, and if s was greater than half-order prior to encoding s. Not accepting additional values excludes signatures which are otherwise valid (with regard to the curve).

Code from secp256k1 which sets "v" (aka recid, less 0x27):
...
    if (recid)
        *recid = (secp256k1_num_cmp(&sig->r, &c->order) >= 0 ? 2 : 0) | (secp256k1_fe_is_odd(&r.y) ? 1 : 0);
...
    if (secp256k1_num_cmp(&sig->s, &c->half_order) > 0) {
        secp256k1_num_sub(&sig->s, &c->order, &sig->s);
        if (recid)
            *recid ^= 1;
    }

And recovery:
    if (recid & 2) {
        secp256k1_num_add(&rx, &rx, &c->order);
        if (secp256k1_num_cmp(&rx, &secp256k1_fe_consts->p) >= 0)
            return 0;
    }
...
    secp256k1_ge_set_xo(&x, &fx, recid & 1);
    if (!secp256k1_ge_is_valid(&x))
        return 0;

Gustav-Simonsson:
@subtly How do you mean? The python code linked validates V such as:
``` python
if self.r < N and self.s < P and self.v >= 27 and self.v <= 28:
```
With is a tighter range than the YP, allowing only 27 and 28 whereas YP allows [27,30].

Use of deterministic ECDSA does not change signature validation rules specified in DSA and ECDSA.

subtly:
@Gustav-Simonsson The link was referring to the comment text specifying valid ranges. The v parameter isn't specified in DSA or ECDSA so it's range wouldn't change those rules.

CJentzsch:
V regarding this PR:
[13:48:10] Vitalik Buterin: however we declare the upper two possibilities, representing infinite values, invalid.
[13:48:14] Vitalik Buterin: that's not true
[13:48:21] Vitalik Buterin: 29, 30 don't represent infinite values
[13:48:40] Vitalik Buterin: we are declaring them invalid, but they represent situations where a coordinate ends up being between N and P

Gustav-Simonsson:
@subtly @CJentzsch @vbuterin Any suggestion for how to update v value? Do we need to? This PR should be unrelated to the definition of the v value, I think. Because the ranges changed here are not depending on the v value, as v is not defined in DSA nor ECDSA specs. It's also not defined in tools.ietf.org/html/rfc6979

gavofyork:
open a new issue for V - i'll merge this as it stands.

PR #117, opened 2015-04-16
Ethash

CJentzsch:
all ethash PR's together. Just for convenience and Proof-reading.


gavofyork:
maybe get @programmerTim or @vbuterins to check this?

CJentzsch:
since it is only one not-merged PR left, we don't need this anymore.

PR #118, opened 2015-04-16
add parentHash formalization

CJentzsch:


Paper.tex
```diff
@@ -362,6 +362,7 @@ \subsubsection{Holistic Validity}
 \begin{array}[t]{lclc}
 H_r &\equiv& \mathtt{\small TRIE}(L_S(\Pi(\boldsymbol{\sigma}, B))) & \wedge \\
 H_o &\equiv& \mathtt{\small KEC}(\mathtt{\small RLP}(L_H^*(B_\mathbf{U}))) & \wedge \\
+H_p &\equiv& \mathtt{\tiny KEC}(\mathtt{\tiny RLP}(P(H))) & \wedge \\
```
gavofyork: `\small KEC` and `\small RLP` like everywhere else?

gavofyork: isn't `P(H)` defined using `H_p`?

CJentzsch:
Yes we define it in sec. 4.3.4 (Block Header Validity). The question is more, if it should be mentioned as part of the holistic validity. But I guess its fine. It is also implicitly used in the block header validity function.

PR #119, opened 2015-04-16
update number of parameters

CJentzsch:
remember the days when the VM was so simple that we needed only 10 parameters.... :-)


gavofyork:
ahh... happy days.

PR #129, opened 2015-05-13
modulo 2^160 for stack items as addresses

CJentzsch:
additionally, µ_r is not defined and not needed. this is handled in eq. 126 in sec 9.4.4 (for STOP opcode)


gavofyork:
rebase?

PR #130, opened 2015-05-18
remove unecessary lack of gas exception

CJentzsch:
If we go out of gas because we can not pay for the CREATE opcode, than there is no return value. We have a OOG exception.
Alternatively, if this meant OOG during the init code execution, than we should rather use the exceptional halting function Z to consider also the other reasons for exceptional halting.


CJentzsch:
[16:18:13] gavofyork: so YP(81) and (82) specify that under any exceptional exit (including both out of gas and stackunderflow) the state should be reverted
[16:19:24] gavofyork: gas for the CREATE opcode is taken before any of the CREATE semantics come into play
[16:19:46] gavofyork: so it cannot mean "not enough gas to pay for the CREATE opcode"
[16:20:38] gavofyork: OOG always and only refers to the sub-code-execution environment, i.e. the code executed during the course of evaluating the CREATE/CALL/CALLCODE instruction

PR #138, opened 2015-06-04
update genesis block

CJentzsch:


gavofyork:
This will change yet...

PR #147, opened 2015-10-15
Correct secp256k1n constant.

aabbbaa:


chriseth:
Previous version seems correct, the constant you added is the order of the field, but the symbol defined here is the order of the group. Or am I missing something here?

aabbbaa:
Sorry, everything is right. I got confused.

PR #149, opened 2016-01-02
Remove redundant clauses.

gavofyork:


efaysal:
A C++ coding of Y,DJ and N is 

https://github.com/efaysal/ETHEREUM_VM_TESTING/blob/master/YP_JumpDest_YD_.cc, 

using a number of opcodes for testing showed that DJ is indeed a subset of Y. 

During my testing I picked some opcodes from known contracts in the blockchain. I was surprised by the following thing, JUMPDEST may present in some opcodes in a consecutive way, as , JUMPDEST JUMPDEST JUMPDEST ...

Also some opcodes having "missing-opcode". A kind of testing and optimization are indeed needed before an opcode reach the blockchain.

PR #152, opened 2016-01-21
careful replacement of SHA3 with KECCAK

ethers:
Initial step for https://github.com/ethereum/EIPs/issues/59

### DIFF
--- a/Paper.tex
+++ b/Paper.tex
@@ -773,7 +773,7 @@ Storage fees have a slightly nuanced behaviour---to incentivise minimisation of 
 %More formally, given an instruction, it is possible to calculate the gas cost of executing it as follows:
 %
 %\begin{itemize}
-%\item {\small SHA3} costs $G_{sha3}$ gas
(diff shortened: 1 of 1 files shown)


gavofyork:
happy to merge this, but i think it needs to happen only following an EIP ratifying the alteration in nomenclature.

ethers:
Solidity now has an alias named `keccak256()`.  Does that help?
https://github.com/ethereum/EIPs/issues/59#issuecomment-252866229

axic:
Shouldn't it be named `KECCAK256` instead?

ethers:
I don't mind either way.  What would you prefer @gavofyork ?

gavofyork:
i think "KECCAK" is fine as long as it's stated as the 256-bit variant somewhere at the beginning. (technically we used SHA3-256, but we always drop the "-256" suffix)

gavofyork:
needs resolving

gavofyork:
this seems to have made it into `master` - closing.

PR #167, opened 2016-03-10
Typos, formatting, and inconsistent definitions

Isweet:
1. Typo in equation (105), the 3rd return value shouldn't be \textbf{s} (which represents suicides), it should be A to represent substate.
2. Also in (105), definition was overlapping with equation label, so added some extra whitespace with \qquad
3. The definition of \Xi expanded the substate in the return, which I think is confusing. I changed it back to A. If you think this is confusing, I think the reader should be redirected back to the definition of substate in section 6.1
4. The definition of O, the stepping function, was defined as taking a tuple, but was called in the definition of X as a function of four arguments. Since O never takes the result of another function which returns a tuple, it makes more sense to define it in function notation.

### DIFF
--- a/Paper.tex
+++ b/Paper.tex
@@ -760,7 +760,7 @@ The account's associated code (identified as the fragment whose Keccak hash is $
 \boldsymbol{\sigma} & \text{if} \quad \boldsymbol{\sigma}^{**} = \varnothing \\
 \boldsymbol{\sigma}^{**} & \text{otherwise}
 \end{cases} \\
-(\boldsymbol{\sigma}^{**}, g', \mathbf{s}, \mathbf{o}) & \equiv & \begin{cases}
+\qquad (\boldsymbol{\sigma}^{**}, g', A, \mathbf{o}) & \equiv & \begin{cases}
 \Xi_{\mathtt{ECREC}}(\boldsymbol{\sigma}_1, g, I) & \text{if} \quad r = 1 \\
 \Xi_{\mathtt{SHA256}}(\boldsymbol{\sigma}_1, g, I) & \text{if} \quad r = 2 \\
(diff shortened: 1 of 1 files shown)


Paper.tex
```diff
@@ -239,7 +239,7 @@ \subsection{The Transaction} \label{ch:transaction}
 \item[gasLimit] A scalar value equal to the maximum amount of gas that should be used in executing this transaction. This is paid up-front, before any computation is done and may not be increased later; formally $T_g$.
 \item[to] The 160-bit address of the message call's recipient or, for a contract creation transaction, $\varnothing$, used here to denote the only member of $\mathbb{B}_0$ ; formally $T_t$.
 \item[value] A scalar value equal to the number of Wei to be transferred to the message call's recipient or, in the case of contract creation, as an endowment to the newly created account; formally $T_v$.
-\item[v, r, s] Values corresponding to the signature of the transaction and used to determine the sender of the transaction; formally $T_w$, $T_r$ and $T_s$. This is expanded in Appendix \ref{app:signing}.
+\item[w, r, s] Values corresponding to the signature of the transaction and used to determine the sender of the transaction; formally $T_w$, $T_r$ and $T_s$. This is expanded in Appendix \ref{app:signing}.
```
subtly: This is not a typo.

gavofyork:
probably good, but @subtly 's comment is correct

axic:
@Isweet can you please rebase and apply the requested changes?

Isweet:
Will take care of this over the weekend.

Isweet:
This PR was pretty outdated w.r.t. the current document, so I created a fresh PR that addresses the issues which were not addressed already (namely, only equation 105 still had one formatting issue and did not refer to A)

Issue #178, opened 2016-07-09
Wrong δ for DELEGATECALL


ralph-pichler:
In the paper (in Appendix H) it says δ=7 for DELEGATECALL. However it seems clients (at least geth and parity) only pop 6 elements off the stack for it.

ethers:
close this @gavofyork ?

pirapira:
close this @gavofyork @chriseth, or you can give me the write permission :cat:

PR #179, opened 2016-07-09
Fix δ for DELEGATECALL

nicksavers:
As mentioned by @ralph-pichler in #178
The DELEGATECALL only pops 6 elements from the stack.


vbgl:
If only six words are popped from the stack, then μ_s[2] no longer represents the value. Therefore, the ninth argument of Θ should be changed to 0 and the condition “μ_s[2] ≤ σ[I_a]_b” should be removed (since it trivially holds).

gavofyork:
agree with @vbgl , though can probably wait for another PR.

PR #201, opened 2016-12-22
Fix the memory consumption of CODECOPY

pirapira:
The argument of the `M` function was a position on the code, but instead it should be the size of the used memory region.


pirapira:
@chriseth I believe this one is ready to merge.

PR #207, opened 2016-12-30
The output from a CALL is considered as the empty string when the call fails

pirapira:
Before this PR, the output was defined as being itself.


pirapira:
Before this PR, the definition of `\boldsymbol{o}` was malformed.  `\boldsymbol{o}` on the left hand side was defined as `\boldsymbol{o}` on the right hand side.  The correct definition seems to be `()` which means the byte sequence of zero-length.

PR #208, opened 2016-12-30
Dealing with returned empty states in CALL instruction

pirapira:
The Θ function returns an empty state on exceptional halts.  In that case the calling `CALL` instruction needs to recover the original state.


gavofyork:
> The Θ function returns an empty state on exceptional halts

what makes you think this? see eq (105)

pirapira:
I don't understand my PR.  I'm not seeing relevant changes since December.  Sorry.

PR #211, opened 2017-01-10
Add the Travis YML file that builds Paper.pdf

pirapira:
This pull-request adds a Travis configuration file.  Travis builds Paper.pdf.  When the sources contain an error, `pdflatex` does not wait for any input but exits abnormally.


pirapira:
@chriseth here is the version that works.

PR #216, opened 2017-01-12
Ethash isn't planned, but is the reality for current Ethereum

axic:


axic:
I wonder if the two references to `Ethereum 1.0` should be made less ambiguous.

PR #222, opened 2017-01-13
Invalid transition

### LINKED ISSUE #90: Extraneous comma
The comma below is terribly out of place.

```
 Invalid state changes might, e.g. be things such as reducing and account balance without an equal and opposite increase elsewhere.
```

pirapira:
Fixes #90.  Also changes the example of an invalid state transition.

The previous example of decreasing the balance without a counter increase, was possible when a contract  executes `SELFDESTRUCT` with its own address.


gavofyork:
the comma is fine, but the example not depends on the storage concept, despite it not having been introduced and not being particularly trivial.

i would say that the example itself is fine - it's meant as a tangible means of illustrating what "invalid" means. it is prefixed with "might" in order to soften it accordingly. i would go as far as to say that in this case, `SUICIDE` is slightly wrong.

pirapira:
Now this PR is just about a comma, and that would fix #90.

PR #230, opened 2017-02-13
Make sure INVALID instruction halts the execution.

pirapira:
Before this pull request, the INVALID instruction did not halt the execution.  The exceptional halting condition (126) detects invalid instrucitons by looking at the value of `\delta`.  Before this pull request, the value of `\delta` for `INVALID` was zero; this did not trigger exceptional halting.  After this pull-request, the value of `\delta` is `\varnothing`, and this triggers exceptional halting.


pirapira:
@chriseth this is a fix to #227, which was merged today.

PR #240, opened 2017-02-20
The execution function X returns remaining gas instead of a state

### LINKED ISSUE #238: Is the second element returned by \Xi a gas or a machine state?
In `master`, we have two equations about the execution function `\Xi`.

```
(σ', g', s, l, r, o) ≡ Ξ(σ,g,I)
```
and
```
Ξ(σ, g, I) ≡ X₀,₁,₂,₄((σ, μ, A⁰, I))
```

According to the first equation, the second element of the return value of `\Xi` is a number.  According to the second, the second element of the return value is a machine state.  This inconsistency needs to be resolved somehow.

pirapira:
Fixes #238


pirapira:
Closing for now.  This "drop the fourth element and take the gas of the second element" operation needs to be defined as a function with some description.

Issue #239, opened 2017-02-20
Different treatment of refunded gas after exceptional halt


pirapira:
In Section 7 "Contract Creation" and in Section 8 "Message Call", the refunded gas is defined differently, especially after exceptional halting.

In Section 7, `g'` is defined in terms of `g**` and the returned global state (indicating if the halting was normal or exceptional.)  However, in Section 8, `g'` is directly taken from the return value of `Ξ`.  I think Section 8 should do the same as Section 7.

I need to check the clients' behaviors, but I'm sure all clients agree because this is a frequent case.

This has to be fixed before #232.  Otherwise the gas-refunding nature of "REVERT" cannot be expressed.

Issue #232, opened 2017-02-13
EIP140: REVERT


pirapira:
This issue keeps track of adoption of https://github.com/ethereum/EIPs/pull/206 <del>and https://github.com/ethereum/EIPs/pull/207/files</del>, which introduce the REVERT instruction.

EDIT: https://github.com/ethereum/EIPs/pull/207 is not about protocol consensus anyway, so not relevant for the Yellow Paper.

pirapira:
* [ ] The sentence "no instruction can, through its instruction, cause an exceptional halt" needs to be changed.

pirapira:
* [ ] `\Xi` and `\Lambda` needs to return gas after REVERT.

pirapira:
#238 and #239 needs to be fixed first.  Otherwise, the gas-returning nature of REVERT cannot be expressed.

msuiche:
@pirapira Can we expect the *REVERT* instruction to be appended to the yellowpaper ?

If not it would be misleading to not have every instruction documented by the yellowpaper, especially when they can be found in the instruction list of the source code of Solidity:
https://github.com/ethereum/solidity/blob/develop/libevmasm/Instruction.cpp#L162

Or is it because it hasn't been merge to `release` branch yet ?

pirapira:
@msuiche there is a pull-request https://github.com/ethereum/yellowpaper/pull/242 about that.

PR #241, opened 2017-02-20
Make sure that message call consumes all gas when exceptional halting…

pirapira:
… happens

This solves #239


pirapira:
@chriseth please have a look because REVERT specification #232  depends on this.

PR #260, opened 2017-03-07
[consensus uncertain] EIP 166: Change the transaction validity condition on the nonce

### LINKED ISSUE #259: [consensus uncertain] EIP166: replay protection using higher bits in the nonce
This issue keeps track of the adoption of https://github.com/ethereum/EIPs/issues/166

pirapira:
If merged, would fix #259


pirapira:
According to https://github.com/ethereum/EIPs/blob/master/README.md, EIP 166 is not considered or accepted.  Closing.

PR #263, opened 2017-03-21
Ξ now returns the gas of the resultant machine state, not the whole resultant machine state

pirapira:
Ξ's output was declared to be (σ', g', A, I) and especially the second element g' was a gas amount while Ξ was defined to return a machine state μ' as the second element.

This PR changes the definition of Ξ so that it returns a gas amount as the second element.

This PR is a reincarnation of #240.

### DIFF
--- a/Paper.tex
+++ b/Paper.tex
@@ -851,7 +851,7 @@ We must now define the $\Xi$ function. In most practical implementations this wi
 The empty sequence, denoted $()$, is not equal to the empty set, denoted $\varnothing$; this is important when interpreting the output of $H$, which evaluates to $\varnothing$ when execution is to continue but a series (potentially empty) when execution should halt.
 \begin{eqnarray}
 \Xi(\boldsymbol{\sigma}, g, I) & \equiv & (\boldsymbol{\sigma}'\!, \boldsymbol{\mu}'_g, A, \mathbf{o}) \\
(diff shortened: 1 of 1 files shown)


Paper.tex
```diff
@@ -850,7 +850,8 @@ \subsection{Execution Overview}
 
 The empty sequence, denoted $()$, is not equal to the empty set, denoted $\varnothing$; this is important when interpreting the output of $H$, which evaluates to $\varnothing$ when execution is to continue but a series (potentially empty) when execution should halt.
 \begin{eqnarray}
-\Xi(\boldsymbol{\sigma}, g, I) & \equiv & X_{0,1,2,4}\big((\boldsymbol{\sigma}, \boldsymbol{\mu}, A^0, I)\big) \\
+\Xi(\boldsymbol{\sigma}, g, I) & \equiv & (\boldsymbol{\sigma}'\!, \boldsymbol{\mu}'_g, A, \mathbf{o}) \\
+(\boldsymbol{\sigma}, \boldsymbol{\mu}'\!, A, I'\!, \mathbf{o}) & \equiv & X\big((\boldsymbol{\sigma}, \boldsymbol{\mu}, A^0\!, I)\big) \\
```
gavofyork: consider `I'\!` -> `...` to make clear that the value isn't being used.

PR #265, opened 2017-03-21
Merge master into Metropolis

pirapira:
... or do you prefer rebasing?


pirapira:
`metropolis` is no longer a planned version.  It has been split into at least two.

PR #258, opened 2017-03-03
EIP 101: big-int precompile

pirapira:
If merged, this would solve #235 as part of the Metropolis changes.

Merging this requires the community's consensus on https://github.com/ethereum/EIPs/issues/101

Issue #235, opened 2017-02-15
EIP 101: big-int precompiles


pirapira:
This issue keeps track of adoption of https://github.com/ethereum/EIPs/issues/101

pirapira:
Since #101 and #198 use different addresses, there is uncertainty about what address is assigned to which contract. I added [WIP] in the title.

pirapira:
This is still WIP because https://github.com/ethereum/EIPs/pull/198 does not yet specify the value of `x % 0`.

Issue #229, opened 2017-02-10
Byzantium changes


pirapira:
We can check these boxes when we create PRs for the yellow paper, or the EIPs are dropped from Metropolis.

* [x] <del>[EIP 5/8](https://github.com/ethereum/EIPs/issues/8)</del> https://github.com/ethereum/EIPs/pull/211 (RETURNDATACOPY and RETURNDATASIZE) | #264 
* [x] [EIP 86](https://github.com/ethereum/EIPs/pull/208) to be followed in #231 (initial abstraction changes) | #277
* [x] [EIP 96](https://github.com/ethereum/EIPs/pull/210) (block hash in a special contract) | https://github.com/ethereum/yellowpaper/pull/311
* [x] [EIP 98](https://github.com/ethereum/EIPs/pull/658) remove medstate from transaction receipts | already in `metropolis` branch
* [x] [EIP 100](https://github.com/ethereum/EIPs/issues/100) (difficulty adjustment) (equation (43).  2017-02-17: two options still; 2017-03-02 perhaps option 1(b)) to be followed in #253 | #266 
* [x] <del>[EIP 101](https://github.com/ethereum/EIPs/issues/101)</del> [EIP 198](https://github.com/ethereum/EIPs/pull/198) (Big-int precompiles) to be followed in #235 | #268 
* [x] <del>[EIP 116](https://github.com/ethereum/EIPs/issues/116)</del> https://github.com/ethereum/EIPs/pull/214 (STATIC_CALL) to be followed in #234 | #270 
* [x] <del>[EIP 166](https://github.com/ethereum/EIPs/issues/166) replay protection with higher-order bits of nonce, to be tracked in #259</del>
* [x] EIP <del>[196](https://github.com/ethereum/EIPs/issues/196)</del> https://github.com/ethereum/EIPs/pull/213 & <del>[197](https://github.com/ethereum/EIPs/issues/197) </del> https://github.com/ethereum/EIPs/pull/212 (zkSNARK verification primitives) | https://github.com/ethereum/yellowpaper/pull/297
* [x] EIP [206](https://github.com/ethereum/EIPs/pull/206) (REVERT) to be followed in https://github.com/ethereum/yellowpaper/issues/232 | #278 
* [x] EIP [649](https://github.com/ethereum/EIPs/pull/669) | https://github.com/ethereum/yellowpaper/pull/333
[...]

PR #268, opened 2017-03-22
Metroplis: EIP 101 big-modexp precompile

pirapira:
If merged, this would solve #235 as part of the Metropolis changes #229.

This is a re-incarnation of #258.  Comments given there are addressed.


axic:
I wonder if it would make sense including the accepted EIP numbers next to the precompiles? Sort of keeping a history which EIP introduced a certain change.

pirapira:
@axic I'll discuss that with @gavofyork when I see him next time.

gavofyork:
i think the EIP process is documentation enough - YP is meant as a clean expression of the current state of the Ethereum protocol, not a chronicle of how it got there.

pirapira:
The commit history should be cleaned, but as a precaution against EIP changes before Metropolis, some intermediate commits are still left.

pirapira:
This PR should now target `master` because now `master` is the Byzantium branch.

PR #276, opened 2017-03-26
Fix stack index error in SELFDESTRUCT

Sword-Smith:
SELFDESTRUCT sends balance to mu_s[0], not to mu_s[1].


pirapira:
Nice catch! :+1:

PR #279, opened 2017-03-30
Insufficient gas to store code results in an exception

pirapira:
This fixes a mistake in the specification of Homestead.

Before this PR, insufficient gas for storing code resulted in underflow in a subtraction.


pirapira:
This problem has been solved on `master`.

PR #283, opened 2017-04-01
Fix remaining issues from PR #167, namely equation 105

Isweet:
This addresses remaining issues of PR #167.

I will close the other PR in favor of this one.

### DIFF
--- a/Paper.tex
+++ b/Paper.tex
@@ -755,7 +755,7 @@ g' & \equiv & \begin{cases}
 0 & \text{if} \quad \boldsymbol{\sigma}^{**} = \varnothing \\
 g^{**} & \text{otherwise}
 \end{cases} \\
-(\boldsymbol{\sigma}^{**}, g^{**}, \mathbf{s}, \mathbf{o}) & \equiv & \begin{cases}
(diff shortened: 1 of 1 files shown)


Paper.tex
```diff
@@ -755,7 +755,7 @@ \section{Message Call} \label{ch:call}
 0 & \text{if} \quad \boldsymbol{\sigma}^{**} = \varnothing \\
 g^{**} & \text{otherwise}
 \end{cases} \\
-(\boldsymbol{\sigma}^{**}, g^{**}, \mathbf{s}, \mathbf{o}) & \equiv & \begin{cases}
+\qquad (\boldsymbol{\sigma}^{**}, g^{**}, A, \mathbf{o}) & \equiv & \begin{cases}
```
pirapira: Does the `\qquad` make a difference here?  I thought the whole tuple is right-aligned toward the `\equiv` symbol.

Isweet: The equation numbering overlaps with the equation. This just scoots things over a bit. If theres a more elegant solution than qquad i'd be happy to do that instead.

pirapira: I see.  I have the same issue sometimes.  Thanks.

pirapira:
Looks good.

PR #285, opened 2017-04-10
Fix typo: T_t should be T_\mathbf{t}

JustinDrake:
Also cleanup nomenclature around byte sequences.


pirapira:
I find this more consistent than before.

PR #297, opened 2017-04-18
Byzantium: elliptic curve precompiled contracts

pirapira:
This pull-request specifies the zkSNARK related precompiled contracts proposed in https://github.com/ethereum/EIPs/pull/213 and https://github.com/ethereum/EIPs/pull/212.

The gas costs in these EIPs are not determined yet, so this is still work in progress.  Also, since the EIPs specify an overlapping addresses for these precompiled contracts, the addresses are still not specified.


pirapira:
This should target `master` because `master` is the Byzantium branch now.

PR #298, opened 2017-04-18
Clarify dependency of H on h in BLOCKHASH instruction

pirapira:
In the description of `BLOCKHASH` instruction, there was a symbol `H` whose value was determined in a somewhat implicit way.  This PR makes it more explicit that `H(h)` is the block header corresponding to its hash `h`.


pirapira:
Rebased.

PR #300, opened 2017-04-20
The nonce in a block header is not a hash.

pirapira:
Before this PR, the nonce of a block header was described to be a "hash" but it is not.  Probably it is a value, so I change it to a value.


pirapira:
@chriseth will you review and merge this?  This is a trivial point but two people asked me about this today.

karalabe:
Take care. I'm not sure it's a value. Afaik it's a 64 bit byte array. It can have leading zeroes. Can it not?

pirapira:
@karalabe I found "the Ethereum address $A(p_r)$ (a 160-bit value)" somewhere in Paper.tex.  So, values seem to contain leading zeroes sometimes.  I don't see "...-bit byte array" anywhere.

nicksavers:
@pirapira @karalabe @chriseth Can you come to an agreement?

pirapira:
@karalabe In the Yellow Paper, a value can contain leading zeroes.  An address is a 160-bit value.

In the Yellow Paper, a byte array is always variable-length.

If we say "8-byte array" here, that would be the first fixed-length byte array in the Yellow Paper.  I think that's not consistent.  For consistency, I prefer "64-bit value".

nicksavers:
If nobody objects, I'll merge it. Just needs a rebase.

nicksavers:
@pirapira I just notice that this change will undo part of this commit https://github.com/ethereum/yellowpaper/commit/c6304c125221f1a1969fa7c6752fdd811ced80c1. It's just about changing the word `hash` to `value` for you right?

pirapira:
You're right.  Fixing.

pirapira:
Done.

PR #303, opened 2017-04-20
After CREATE, the 1/64 of the original available gas is added again

pirapira:
Before this commit, after a successful CREATE, the available gas decreased at least by 1/64.
This problem was introduced in EIP-150 changes.  The gas passed on was reduced by 1/64 but this 1/64 was then forgotten.


pirapira:
This one should target Tangerine Whistle.

pirapira:
To be reopened on `tangerie-whistle`.

PR #304, opened 2017-04-20
Gas usage of CALL should first take C_CALLGAS into account

pirapira:
Before this change, the CALL instruction was receiving the remainder of the gas from the callee, but not losing the initial gas passed to the callee.
This change makes the caller to lose the initial gas passed to the callee before receiving the remainder.


pirapira:
Maybe this PR should target `frontier`.

pirapira:
I'll reopen this on `frontier`.

PR #305, opened 2017-04-25
ECRECOVER precompiled contract should have stayed unaffected by EIP155 changes

pirapira:
~Before this PR, the definition of ECREC precompiled contract seemed to have changed during Homestead changes.  The Homestead changes involved stricter checks on the signature, and on the Yellow Paper, that change also affected the recovering precompiled contract.  However, the precompiled contract did not change in the implementations.~  I think it's about EIP155 replay protection.  EIP155 should not change the ECDSARECOVER precompiled contract.

### DIFF
--- a/Paper.tex
+++ b/Paper.tex
@@ -1705,15 +1705,15 @@ We assume the existence of functions $\mathtt{\small ECDSAPUBKEY}$, $\mathtt{\sm
 \mathtt{\small ECDSARECOVER}(e \in \mathbb{B}_{32}, v \in \mathbb{B}_{1}, r \in \mathbb{B}_{32}, s \in \mathbb{B}_{32}) & \equiv & p_{\mathrm{u}} \in \mathbb{B}_{64}
 \end{eqnarray}
 
-Where $p_{\mathrm{u}}$ is the public key, assumed to be a byte array of size 64 (formed from the concatenation of two positive integers each $< 2^{256}$), $p_{\mathrm{r}}$ is the private key, a byte array of size 32 (or a single positive integer in the aforementioned range) and $e$ is the hash of the transaction, \hyperlink{h_of_T}{$h(T)$}. It is assumed that \hypertarget{v}{}$v$ is either the `recovery identifier' or `chain identifier doubled plus 35 or 36'. The recovery identifier is a 1 byte value specifying the parity and finiteness of the coordinates of the curve point for which $r$ is the x-value; this value is in the range of $[27, 30]$, however we declare the upper two possibilities, representing infinite values, invalid. The value 27 represents an even $y$ value and 28 represents an odd $y$ value. In the second case, where \hypertarget{v}{}$v$ is the chain identifier doubled plus 35 or 36, the values 35 and 36 assume the role of the `recovery identifier' by specifying the parity of $y$, with the value 35 representing an even value and 36 representing an odd value.
+Where $p_{\mathrm{u}}$ is the public key, assumed to be a byte array of size 64 (formed from the concatenation of two positive integers each $< 2^{256}$), $p_{\mathrm{r}}$ is the private key, a byte array of size 32 (or a single positive integer in the aforementioned range) and $e$ is the hash of the transaction, \hyperlink{h_of_T}{$h(T)$}. It is assumed that \hypertarget{v}{}$v$ is the `recovery identifier'. The recovery identifier is a 1 byte value specifying the parity and finiteness of the coordinates of the curve point for which $r$ is the x-value; this value is in the range of $[27, 30]$, however we declare the upper two possibilities, representing infinite values, invalid. The value 27 represents an even $y$ value and 28 represents an odd $y$ value.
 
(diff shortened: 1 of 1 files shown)


pirapira:
Maybe this should be merged into `homestead` and then merged into more recent versions.

Paper.tex
```diff
@@ -1418,6 +1418,13 @@ \section{Signing Transactions}\label{app:signing}
 %\mathtt{\tiny secp256k1p} &= 2^{256} - 2^{32} - 977\\
 \end{align}
 
+Note that $\mathtt{\small ECDSARECOVER}$ function does not follow the distinction of valid or invalid transactions defined here. $\mathtt{\small ECDSARECOVER}$ returns non-empty results if the following conditions are met (this is relevant in the definition of $\Xi_{\mathtt{ECREC}}$).
```
jamesray1: Note that the ECDSARECOVER function.

Paper.tex
```diff
@@ -1418,6 +1418,13 @@ \section{Signing Transactions}\label{app:signing}
...
+\begin{align}
+0 < r &< \mathtt{\tiny secp256k1n} \\
+0 < s &< \mathtt{\tiny secp256k1n} \\
```
jamesray1: Get rid of tiny.

Paper.tex
```diff
@@ -1418,6 +1418,13 @@ \section{Signing Transactions}\label{app:signing}
...
+v &\in \{27,28\}
```
jamesray1: This conflicts with EIP-155 and https://ethereum.github.io/yellowpaper/paper.pdf#v.

jamesray1: You may also notice "These are formally defined in theliterature, e.g. by ?." #630 should fix that.

jamesray1: Would be good to add to the build script to pick up Latex errors.

jamesray1: I think I understand the conflict. Following are the two definitions of an invalid signature.

screenshot from 2018-02-21 19-28-07

screenshot from 2018-02-21 19-29-26

jamesray1: Wouldn't it also be good to update the implementations and the first of the above?

jamesray1: Noticed something else: #632.

jamesray1: Wouldn't it be better to update the EIP draft, implementations and then this paper for [the first screenshot with the ECREC precompile](https://ethereum.github.io/yellowpaper/paper.pdf#page.20) so that an invalid signature is defined in accordance with the second screenshot in [Appendix F](https://ethereum.github.io/yellowpaper/paper.pdf#appendix.F)?

pirapira: That will be a hardfork, so, there needs to be a separate EIP proposing the fix.

jamesray1: #893

Paper.tex
pirapira: What do you mean?

jamesray1: These are proposed edits.

Note that *the* $\mathtt{\small ECDSARECOVER}$ function does not follow the distinction of valid or invalid transactions defined *above*. $\mathtt{\small ECDSARECOVER}$ returns non-empty results if the following conditions are met (this is relevant in the definition of *the precompile* $\Xi_{\mathtt{ECREC}}$).

ldct:
LGTM. It is unfortunate that 2 different conditions for valid ECDSA verification need to be used.

Paper.tex
pirapira: That looks better.  I'll modify my branch.

pirapira:
@jamesray1 fixed my wording.

pirapira:
@nicksavers pointed out that, this PR explains ECRECOVER precompiled as some modification of transaction signing, but it should be the other way around. The transaction signature checking (considering replay attack prevention) should be a more complicated variant of ECRECOVER operation (that's used in the ECRECOVER precompiled). I agree with this pereference.

pirapira:
Since YP delegates the definition of ECRECOVER to a paper by Johnson, Menezes and Vanstone, I have to see first which `v` does the original `ECRECOVER` accept.

pirapira:
But this document explains signature generation and verification, not the recovery of the public key.  I created an issue about that https://github.com/ethereum/yellowpaper/issues/688 .  For this PR I assume that the original ECDSA doesn't care about the chain IDs.

pirapira:
Some sentences on https://github.com/ethereum/yellowpaper/pull/305/files#diff-9f702e1491c55da9d76a68d651278764R1703 talk about chain_ids, these should be moved downwards to around https://github.com/ethereum/yellowpaper/pull/305/files#diff-9f702e1491c55da9d76a68d651278764R1761

pirapira:
And also, the values except 27 and 28 should not appear https://github.com/ethereum/yellowpaper/pull/305/files#diff-9f702e1491c55da9d76a68d651278764R1711.  They have to wait until we see signed transactions.

pirapira:
@nicksavers this should be better now.

pirapira:
Ah wait, I think this is still somehow wrong.  So the precompiled contract should not check the value of `v` so strictly...

PR #306, opened 2017-04-25
Pull ommer's uniqueness test out of the "is-sibling" equation.

dstarcev:
The test, verifying that an ommer is included into a block only once was defined in the "is-sibling" equation, which is considered a wrong place for this test.

### DIFF
--- a/Paper.tex
+++ b/Paper.tex
@@ -985,7 +985,7 @@ The process of finalising a block involves four stages:
 
 \subsection{Ommer Validation}
 
-The validation of ommer headers means nothing more than verifying that each ommer header is both a valid header and satisfies the relation of $N$th-generation ommer to the present block where $N \leq 6$. The maximum of ommer headers is two. Formally:
+The validation of ommer headers means nothing more than verifying that each ommer header is both a valid header and satisfies the relation of $N$th-generation ommer to the present block where $N \leq 6$. The maximum of ommer headers is two. An ommer may be included only in one block in a blockchain. Formally:
 \begin{equation}
 \lVert B_\mathbf{U} \rVert \leqslant 2 \bigwedge_{U \in B_\mathbf{U}} V(U) \; \wedge \; k(U, P(B_H)_H, 6)
 \end{equation}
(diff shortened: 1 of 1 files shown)


Paper.tex
```diff
@@ -985,22 +985,23 @@ \section{Block Finalisation} \label{ch:finalisation}
 
 \subsection{Ommer Validation}
 
-The validation of ommer headers means nothing more than verifying that each ommer header is both a valid header and satisfies the relation of $N$th-generation ommer to the present block where $N \leq 6$. The maximum of ommer headers is two. Formally:
+The validation of ommer headers means nothing more than verifying that each ommer header is both a valid header and satisfies the relation of $N$th-generation ommer to the present block where $N \leq 6$. The maximum of ommer headers is two. An ommer may be included only in one block in a blockchain. Formally:
```
pirapira: Is there a list of all ommers that have been included in any block in a blockchain?  How is it enforced?

jamesray1: So should this PR be closed?

Paper.tex
```diff
@@ -985,22 +985,23 @@ \section{Block Finalisation} \label{ch:finalisation}
...
+(s(U, H) &\\
+\quad \vee \; k(U, P(H)_H, n - 1)) &\\
+\quad \wedge \; U \notin B(H)_\mathbf{U} & \text{otherwise}
 \end{cases}
 \end{equation}
 
 and $s$ denotes the ``is-sibling'' property:
 \begin{equation}
-s(U, H) \equiv (P(H) = P(U)\; \wedge \; H \neq U \; \wedge \; U \notin B(H)_\mathbf{U})
+s(U, H) \equiv (P(H) = P(U)\; \wedge \; H \neq U)
```
jamesray1: In order to be a sibling you need to rule out that they are not in the same height in the chain, thus AIUI why we have `\; \wedge \; U \notin B(H)_\mathbf{U})` here. Moving it to the is-kin property doesn't make sense.

Paper.tex
nicksavers: An ommer (uncle) is a block header with a `parentHash` and it enforced by comparing this with the block's parent hash.
https://github.com/ethereum/go-ethereum/blob/master/consensus/ethash/consensus.go#L211

PR #308, opened 2017-04-28
Metropolis: EIP98 option 1

pirapira:
Before this PR, the metropolis branch contained EIP98 option 2, but the consensus now is option 1.


pirapira:
https://github.com/ethereum/EIPs/pull/658/files supersedes this.

PR #313, opened 2017-06-21
Make sure that divisions and modulo operations are not subject to mod 2^{256}

pirapira:


pirapira:
Maybe this one should target `frontier` and merged into newer versions.

pirapira:
I'll reopen this on `frontier`.

PR #316, opened 2017-06-27
make bg white

aakilfernandes:
It eats up a bunch of ink with the yellow bg. Can we remove the bg?


gavofyork:
you certainly can. locally :).

PR #318, opened 2017-07-17
Byzantium: EIP-658 Embedding transaction status code in receipts

pirapira:
EIP: https://github.com/ethereum/EIPs/pull/658

Now transaction receipts contain one byte indicating success or failure.


pirapira:
The target branch should now be `master`.

PR #323, opened 2017-08-04
Contract creation on non-empty code or non-zero nonce throws an exception

pirapira:
This implements a protocol change which is very hard to hit anyway.  The idea is to avoid overwriting contracts even when the calculated address coincides with existing something.  Here "existing something" is an account with non-zero nonce or non-empty code.  This is consistent with EIP86 planned in Constantinople.

There have been some test failures happening only on the address-collision cases, but most probably the time spent there is wasted because keccak256 collisions cause far bigger problems anyway.  It makes sense to make the protocol specification simpler for the address collision case.

### DIFF
--- a/Paper.tex
+++ b/Paper.tex
@@ -760,7 +760,7 @@ g^{**} - c & \text{otherwise} \\
 1 & \text{otherwise}
 \end{cases} \\
 \nonumber \text{where} \\
-F &\equiv \big((\boldsymbol{\sigma}^{**} = \varnothing \ \wedge\ \mathbf{o} = \varnothing) \vee\  g^{**} < c \ \vee\  |\mathbf{o}| > 24576\big)
+F \equiv &\big(\boldsymbol{\sigma}[a]_c \neq \texttt{\small KEC}\big(()\big) \ \vee\ \boldsymbol{\sigma}[a]_n \neq 0 \ \vee\ (\boldsymbol{\sigma}^{**} = \varnothing \ \wedge\ \mathbf{o} = \varnothing) \\\nonumber  & \vee\ g^{**} < c \ \vee\  |\mathbf{o}| > 24576\big)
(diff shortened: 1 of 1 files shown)


pirapira:
This implements https://github.com/ethereum/EIPs/issues/684

fulldecent:
Is there any more discussion on this. This is something I wondered too when reading the spec.

Paper.tex
```diff
@@ -760,7 +760,7 @@ \section{Contract Creation}\label{ch:create}\hypertarget{endow}{}
 1 & \text{otherwise}
 \end{cases} \\
 \nonumber \text{where} \\
-F &\equiv \big((\boldsymbol{\sigma}^{**} = \varnothing \ \wedge\ \mathbf{o} = \varnothing) \vee\  g^{**} < c \ \vee\  |\mathbf{o}| > 24576\big)
+F \equiv &\big(\boldsymbol{\sigma}[a]_c \neq \texttt{\small KEC}\big(()\big) \ \vee\ \boldsymbol{\sigma}[a]_n \neq 0 \ \vee\ (\boldsymbol{\sigma}^{**} = \varnothing \ \wedge\ \mathbf{o} = \varnothing) \\\nonumber  & \vee\ g^{**} < c \ \vee\  |\mathbf{o}| > 24576\big)
```
livnev: I believe this should be merged. I know that the yellow paper hasn't caught up to Constantinople yet but this would be a start.

ytrezq:
@pirapira since this is now activated on mainnet, maybe the tag is no longer relavant., isn t it?

yperbasis:
For reference: [geth implementation](https://github.com/ethereum/go-ethereum/blob/v1.10.3/core/vm/evm.go#L439)

yperbasis (approved):

PR #327, opened 2017-08-10
URLs

jamesray1:
For archiving purposes, I've added live URLs and archived URLS (where possible) for all references. While I understand that including URLS is not usually a standard style for non-web texts, I think it is good to include them to make it easier to access the references and provide an archive URL to provide a citation that will be accessible for perpetuity.

### DIFF
--- a/Biblio.bib
+++ b/Biblio.bib
@@ -54,7 +54,7 @@
 	note = {{https://web.archive.org/web/20170810035057/https://www.iacr.org/archive/ches2004/31560117/31560117.pdf}},
 	BookTitle = {{Cryptographic Hardware and Embedded Systems-CHES 2004}},
 	Publisher = {{Springer}},
-	author = {Gura, Nils and Patel, Arun and Wander, Arvinderpal and Eberle, Hans and Shantz, Sheueling Chang},
+	author = {Gura, Nils and Patel, Arun and Wander, Arvinderpal and Eberle, Hans and Sheueling Chang Shantz},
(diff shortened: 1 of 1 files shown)


jamesray1:
> Appendix F. Signing Transactions
> The method of signing transactions is similar to the ‘Electrum style signatures’; it utilises the SECP-256k1 curve as described by Gura et al. [2004].

CTRL+F for "256k1" in this referenced paper yields no results. The paper does talk about their implementation strategy on p. 9 of 15, para. 3, perhaps the citation is referring to this?

In this PR, I've added references for secp-256k1 with secp256k1BitcoinWiki2016 and secp256k1StackExchange2014. Once this PR is approved, the references can then be cited in paper.tex.

jamesray1:
Also, ‘Electrum style signatures’ should have a reference, e.g. https://www.npmjs.com/package/electrum, which I've added with npmElectrum2017

jamesray1:
For "we assert the functions ECDSASIGN, ECDSARESTORE and ECDSAPUBKEY. These are formally defined in the literature." there should be references for this. I've added references for [this](http://cs.ucsb.edu/~koc/ccs130h/notes/ecdsa-cert.pdf) and [this](https://en.wikipedia.org/wiki/Elliptic_Curve_Digital_Signature_Algorithm), which have the label ECDSAcerticom and ECDSAWikipedia, respectively. For the first link, refer to section 6.2 for ECDSAPUBKEY, and section 7 for ECDSASIGN and ECDSARECOVER (assuming that the former is for signature generation and the latter is for signature verification).

jamesray1:
Merged into #345.

PR #333, opened 2017-09-08
Byzantium: EIP-649 Difficulty Bomb Delay and Block Reward Reduction

pirapira:
Implements https://github.com/ethereum/EIPs/pull/669/files


pirapira:
This has to target `master`.

PR #336, opened 2017-09-12
minor equation typos

4000D:
103) account state is defined as RLP(nonce, balance, stateRoot, codeHash), but written as (balance, nonce, codeHash, stateRoot)

232) parentheses needed to multiply before subtraction done


pirapira:
@gavofyork I think you can merge this.

PR #348, opened 2017-09-19
add   Chinese version

cuiguangbin:
add Chinese version


nicksavers:
Looks like a nice effort. Although it seems to be only partially done.
Are there people that can independently verify that the transaction was done correctly?

Having said that, I don't think it's smart to include this translation into the same repository because it would take double the effort to keep everything in sync. But now that the paper is licensed under CC BY-SA 4.0, you can publish this yourself.

pirapira:
I'm with @nicksavers .  You can host the Chinese version on your own.  Maybe we can add a list of translations in the README file.

PR #375, opened 2017-10-16
Revert 2 master

jamesray1:


jamesray1:
I'm just trying to change my local master so that's it's even with the Ethereum master.

PR #365, opened 2017-10-01
[WIP] To be rebased into smaller PRs: Improving readability and adding citations: hyperref, cross-references, extra citations; clarifications (e.g. minor rewording and footnotes); list of maths symbols; fixed formatting e.g. subscripts to \mathrm font; archive links; grammar fixes (e.g. complete sentences like in the tables); readme: omitted yellowpaper.io and added more details on how to build.

jamesray1:
Too big a PR, rebasing to smaller PRs:
- [x] PR #373: rebased commits 2 (97b534cd876111602e51eaf3d56864807096ccc4) and 3 (e71e30e58c451724929bd33d2da88608b26923c7), and implicitly commit 1.
- [ ] ~~PR #371: Merged 59bcf73, 51b0afb3393bce05c7143b60cb80218dc757f56d, b2f7605, 08172f4 and 9f03f5896ba66c4cf27658cf4545988a020fdc53 and 60ea6865340e20b68cd648f941d4e9941886d0fc to local Clarifications branch, but got mixed up with other commits. All commits: numbers 4 (08172f4) to 17 (9f03f58). It's probably better to rebase into still smaller PRs.~~
- [ ] ~~PR #370: Merged commits 7 (9072ca8bf5b2205b296fb6de2d04a279b48e3983) and 11, (3feff95f17e7985e8c0b5c1a5dba51e80b608894), and implicitly commit 10 (4710b8e), which was overwritten by 11. Done on Github, but another commit needs to be merge 1e7849255f97ad45b8f5249399cb95f84f782301~~
- [ ] rebase other commits

PR #377, opened 2017-10-16
[WIP] hyperref package

jamesray1:


jamesray1:
I have gone and added more commits to one PR rather than reduced them!

@pirapira is there any chance that you could review by adding comments to the diff and checkmarks of where you are up to? That would be a lot less time-consuming for me.

pirapira:
@jamesray1 At the end, I cannot merge this, but I need to convince @gavofyork .  I don't know his limit of complexity per PR.

jamesray1:
This PR is not suitable for merging, the other one #365 should be more suitable.

jamesray1:
It's just really tedious to have to go through all the commits, split them into separate branches and create separate PRs. There are a lot of commits that have independent changes, but then there are also a lot of cross-dependencies between commits, particularly with the hyperref package.

jamesray1:
It also seems harder to work with Git when there are many branches.

Every time I make a change in Git to keep PRs short it'd be best to create the change in a separate branch with `git checkout` to the branch to duplicate, then `git checkout -b <describe-the-change>` then make the changes locally, then run Git push. It's easier to use Github, but Github isn't good when you need to cherry-pick commits.

jamesray1:
Yoichi, I have a suggestion. Post a link in the readme to a comment that contains a link to my latest version of the paper, with a warning that this version has not been reviewed, approved, or merged. @pirapira

pirapira:
@jamesray1 I cannot do that for the copyright reasons.

jamesray1:
Even if my PR is CC0?

pirapira:
@jamesray1 the original is not CC0.

jamesray1:
Ah I see what you mean there is no license so by default it's all rights reserved.

Issue #317, opened 2017-07-13
Minimum gas limit different on Ethereum clients?


ntallar:
[Equation (53)](https://github.com/ethereum/yellowpaper/blob/master/Paper.tex#L466) refers to the minimum gas limit for blocks as 125000. Should this value be changed to 5000 (0x1388)? Various Ethereum clients seem to be using that minimum gas limit value:
* *Geth*: https://github.com/ethereum/go-ethereum/blob/master/params/protocol_params.go#L68
* *Parity*: https://github.com/paritytech/parity/blob/master/ethcore/res/ethereum/foundation.json#L146

pirapira:
* cpp-ethereum https://github.com/ethereum/cpp-ethereum/blob/develop/libethereum/GenesisInfo.cpp#L36

also uses 5000.

nicksavers:
This seems to have been accidentally changed on 16 Feb 2016 in the commit https://github.com/ethereum/yellowpaper/commit/a3381de9e1a20123cddbe85c9d5cc4f01f00d648
The `minGasLimit` was 5000 at Frontier and hasn't changed in any of the hardforks.

PR #393, opened 2017-11-18
Fix minGasLimit to 5000

nicksavers:
Fix according to https://github.com/ethereum/yellowpaper/issues/317
Major clients implement a minGasLimit of 5000 instead of 125000 as specified in the paper

https://github.com/ethereum/go-ethereum/blob/master/params/protocol_params.go#L76
https://github.com/ethereum/cpp-ethereum/blob/develop/libethereum/GenesisInfo.cpp#L36
https://github.com/paritytech/parity/blob/master/ethcore/res/ethereum/foundation.json#L147
https://github.com/ethereum/ethereumj/blob/develop/ethereumj-core/src/main/java/org/ethereum/config/blockchain/FrontierConfig.java#L44


veox:
You may want to add `Closes #317` (or similar) to the description. ;)

nicksavers:
I noticed another occurrence that I missed 
https://github.com/ethereum/yellowpaper/blob/master/Paper.tex#L493

PR #404, opened 2018-01-08
Create Byzantium-Hard-Fork-changes.md

jamesray1:


nicksavers:
For the fork changes, there are these [Informational EIPs](https://github.com/ethereum/EIPs#past-hard-forks). The yellowpaper repository itself can be kept concise. For every fork there will be a branch or tags, so if necessary they can be compared using a diff.

PR #282, opened 2017-03-31
Spurious Dragon: EIP155 replay protection

pirapira:
This follows https://github.com/ethereum/EIPs/issues/155 as part of the Spurious Dragon changes.

Since I've branched `homestead`, and this change has already been taken into the mainnet, this PR is on `master`.

PR #405, opened 2018-01-13
Improve explanation of "recovery id" used in ECDSARECOVER

Sword-Smith:
r is the calculated x-coordinate  of the "curve point" in the ECDSASIGN function.

Given an x-coordinate, two possible curve points exist. One with a negative y-value
and one with a positive y-value. The recovery id specifies which of the two possible y-values
is the correct one.


nicksavers:
Thanks, I will merge this but not yet as it conflicts with another PR https://github.com/ethereum/yellowpaper/pull/282/ for [EIP-155](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-155.md).
After that is merged, this can be added to clarify the the paragraph.

Sword-Smith:
Thank you for your quick response. I will attempt to commit a better explanation, though. v is determined by parity; I am unsure how much it makes sense to talk about signs when finding square roots in modular arithmetic.

PR #406, opened 2018-01-16
Add table of versions in BRANCHES.md

pirapira:
The repository is organized so that every protocol version has a branch.  This PR documents those branches in README.md.


pirapira:
Perhaps I should add block numbers.

nicksavers:
Perhaps this table is better suited for a separate file called ` BRANCHES`, with a link to it from the README. And yes, including block-numbers would help understand the order in which to compare the branches and implementing a client that supports the full history.

pirapira:
I separated the branch table into a separate file.  I also added the block numbers in the table.

jamesray1:
May we please also put links to each version of the PDF?

jamesray1:
I'm confused. https://github.com/ethereum/EIPs/blob/master/EIPS/eip-606.md says that the fork block number for Homestead for the main net is >=1,150,000, not 1 as is stated here.

pirapira:
@jamesray1 please open a new issue about a link to each version of PDF.

pirapira:
@jamesray1 I think I forgot `frontier` in the table.

PR #407, opened 2018-01-16
Update Paper.tex

Sword-Smith:
r is the calculated x-coordinate of the "curve point" in the ECDSASIGN function.

Given an x-coordinate, two possible curve points exist. One with an odd y-value
and one with an even y-value. The recovery id specifies which of the two possible y-values
is the correct one.

### DIFF
--- a/Paper.tex
+++ b/Paper.tex
@@ -1404,7 +1404,7 @@ We assert the functions $\mathtt{\small ECDSASIGN}$, $\mathtt{\small ECDSARECOVE
 \mathtt{\small ECDSARECOVER}(e \in \mathbb{B}_{32}, v \in \mathbb{B}_{1}, r \in \mathbb{B}_{32}, s \in \mathbb{B}_{32}) & \equiv & p_u \in \mathbb{B}_{64}
 \end{eqnarray}
 
-Where $p_u$ is the public key, assumed to be a byte array of size 64 (formed from the concatenation of two positive integers each $< 2^{256}$) and $p_r$ is the private key, a byte array of size 32 (or a single positive integer in the aforementioned range). It is assumed that $v$ is the `recovery id', a 1 byte value specifying the sign and finiteness of the curve point; this value is in the range of $[27, 30]$, however we declare the upper two possibilities, representing infinite values, invalid.
+Where $p_u$ is the public key, assumed to be a byte array of size 64 (formed from the concatenation of two positive integers each $< 2^{256}$) and $p_r$ is the private key, a byte array of size 32 (or a single positive integer in the aforementioned range). It is assumed that $v$ is the `recovery id', a 1 byte value specifying the parity and finiteness of the coordinates of the curve point for which $r$ is the x-value; this value is in the range of $[27, 30]$, however we declare the upper two possibilities, representing infinite values, invalid. The value 27 represents an even $y$ value and 28 represents an odd $y$ value.
 
 \newcommand{\slimit}{\ensuremath{\text{s-limit}}}
 


Paper.tex
```diff
@@ -1404,7 +1404,7 @@ \section{Signing Transactions}\label{app:signing}
...
 \end{eqnarray}
 
-Where $p_u$ is the public key, assumed to be a byte array of size 64 (formed from the concatenation of two positive integers each $< 2^{256}$) and $p_r$ is the private key, a byte array of size 32 (or a single positive integer in the aforementioned range). It is assumed that $v$ is the `recovery id', a 1 byte value specifying the sign and finiteness of the curve point; this value is in the range of $[27, 30]$, however we declare the upper two possibilities, representing infinite values, invalid.
+Where $p_u$ is the public key, assumed to be a byte array of size 64 (formed from the concatenation of two positive integers each $< 2^{256}$) and $p_r$ is the private key, a byte array of size 32 (or a single positive integer in the aforementioned range). It is assumed that $v$ is the `recovery id', a 1 byte value specifying the parity and finiteness of the coordinates of the curve point for which $r$ is the x-value; this value is in the range of $[27, 30]$, however we declare the upper two possibilities, representing infinite values, invalid. The value 27 represents an even $y$ value and 28 represents an odd $y$ value.
```
pirapira: The new version reads better.  How can I be sure that 27 represents even and 28 represents odd (somehow counterintuitive to me)?

Sword-Smith: @pirapira I will try to argue my claim from a code perspective and from a math perspective.

##Code
I can show you that this is how ethereum-client-haskell handles v.
You can look at the call to "getPubKeyFromSignature (ExtendedSignature (Signature r s) (v == 28)) h" in this file (line 33)
https://github.com/jamshidh/ethereum-vm/blob/master/src/Blockchain/VM/PrecompiledContracts.hs

And then look at the naming of the boolean value which is set to v == 28 in this file. The name is "isOdd" (line 85).
https://github.com/jamshidh/ethereum-encryption/blob/master/src/Blockchain/ExtendedECDSA.hs

## Math
https://etherscan.io/getRawTx?tx=0x118888f8fe461a750a9b0d25a391e552a69a975cb002896ba54af1669896f5db
Furthermore, we can look at this transaction which was included in block 4925095. It uses the pre-EIP155 signature standard (so not EIP155). The v-bit is the 3rd to the last field in the RLP structure. It is 0x1c (followed by the signature fields "a033ef4bbbdefa ...".

*T_w = 0x1c = 28 => v = 1 => y-value for curve point (y-value corresponding to r-value) is odd.*

The x-value of the y-value is The T_r field. That is the second to last is the RLP structure. It is:
0x33ef4bbbdefaf37397cc154f9e567e9764d4fcb406f6b9e4174cc5c97bb2b65d =
23490754422197219708049421328749734613450349085537272446202174922432628045405

Solving the secp256k1 equation ´y^2 = x^3 + 7 mod p´ for this x value yields two y-values:

38002010198178446753887515983660381251721238641880372567923363885670070868102, 77790079039137748669683469025027526601548746023760191471534220122238763803561

I claim that the 2nd square root (the odd one) is the one that was used in this signature. 

So the "curve point" (R) is:
$R =
 (23490754422197219708049421328749734613450349085537272446202174922432628045405, 77790079039137748669683469025027526601548746023760191471534220122238763803561)$

This statement can be checked by performing the calculating $r^(−1)(s*R−z*G)$  in elliptic curve math. The z-value is the hash of the unsigned transaction (remove last three RLP fields from the signed transaction, adjust length bytes, and take KECCAK3). It is:
$z = 9240012045807776463989548829895232747436327928927690755365151397253009812739$

I get the following values:
$r^(-1) = 21489769294208500149889596789439848541111199598010131160088374732782096411659$
$z*G = (109504647668550041297831566283061404004164449100228615705961346764579760993911, 9369131090125488860538804874026358411701863995318542053193668775613697178709)$
$s*R = (40134963095822281995939060545756826093459838722635818059424880664519732503294, 46894798648849248147903245365422854177801872249629258568856068187986863163921)$

This gives the public key:
$pubkey = (92468855070381551177014661507802518244444910835268705198880145840647352756385, 9337666946427467771927585216698334962158487605858241584075198076947056800522)$
In hex:
$pubkey = cc6f846bbda7091734bd6f389577d878e5cc594892a62a55c252b66b63c628a114a4eea4b524e0b27c7becf33e9cc34602562bc56a8ef72be83bc34a5bcfb30a$

Taking KECCAK256 of the public key and selecting the last 160 bits of this value yields the expected address:
0xd72fbae4b64079edd64969f760cdf240ec7533eb

Which matches the address from etherscan.io
0xd72fbae4b64079edd64969f760cdf240ec7533eb

See
https://etherscan.io/tx/0x118888f8fe461a750a9b0d25a391e552a69a975cb002896ba54af1669896f5db

Sword-Smith:
Some resources if anybody wants to go into the math:
https://crypto.stackexchange.com/questions/18105/how-does-recovering-the-public-key-from-an-ecdsa-signature-work
https://crypto.stackexchange.com/questions/11316/subtracting-a-point-in-elliptic-curve-cryptography

Paper.tex
Sword-Smith: The wrong pubkey that we can derive if we use the 
even y-value is: (2492467599879086090592534386748770237491359718291635995047644463732266573281, 78794639673714625778473280360348409184287073268398429254096496503608379247185)
= 0x0582afc984c36063cd51a95a7545c9c487e60ea9a9a06b61b67c04452e9b1de1ae342f85307d703540ce92dbcc6837193acd1b91954fcb902cb7f004dc650a51

This hashes to the address:
0x95d66cb83d243afc016c6eb2958f6cbb475d95d8

Which does not match that on etherscan.io

PR #270, opened 2017-03-23
Metropolis: static call

pirapira:
This PR introduces https://github.com/ethereum/EIPs/pull/214 and fixes #234 as part of the Metropolis changes.

This is a reincarnation of #237.

PR #408, opened 2018-01-17
Static call

pirapira:
This implements [EIP-214](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-214.md).

This is a reincarnation of #270.


pirapira:
@nicksavers pointed out to me that this PR contains EIP-98.  I'll separate that out.

pirapira:
Done.

PR #410, opened 2018-01-17
Byzantium: EIP-101 expmod precompiled contract

pirapira:
If merged, this would solve #235 as part of the Byzantium changes #229.

This is a re-incarnation of #268.


nicksavers:
These commit messages are wrong, but the content is for EIP-198 and **not** EIP-101. Merged nonetheless.

PR #414, opened 2018-01-18
add Chinese version link  in ReadMe.md

yuange1024:
add Chinese version link in ReadMe.md


pirapira:
@yuange1024 by the way, the unit "Wei" comes from 戴维 https://en.bitcoin.it/wiki/Wei_Dai.

PR #422, opened 2018-01-19
May as well comment out \lightpink if it's unused for performance reasons

jamesray1:


pirapira:
Let's remove the line.  Maybe there was a last-moment decision between pink or yellow, who knows.

jamesray1:
Done.

PR #423, opened 2018-01-19
with Bitcoin being one of the most notable ones

jamesray1:


jamesray1:
Oh, I didn't see it that way. I will look at some uses of the two phrases.

To me, not least just sounds like a contraction of not the least, so I just had a grammar correction in mind. Other than that I don't think there is a different semantic meaning to the two phrases.

jamesray1:
I think they mean the same thing.

https://www.merriam-webster.com/dictionary/not%20least

However, "not the least" could be misconstrued with "not in the the least".

https://www.merriam-webster.com/dictionary/not%20the%20least

I think it's better to change it to "particularly".

jamesray1:
Sorry, I forgot to update the commit title.

I used "with Bitcoin being one of the most notable ones".

PR #424, opened 2018-01-19
EIP98: remove state root from transaction receipts

pirapira:


nicksavers:
The new status field is described in [EIP#658](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-658.md).

PR #435, opened 2018-01-20
Update explanations of \homesteadmod and \expdiffsymb with citations for EIPs 2, 100 and 649

jamesray1:


Biblio.bib
```diff
@@ -1,3 +1,26 @@
...
+@Misc{EIP-649,
+	url = "https://github.com/ethereum/EIPs/blob/63c93eb71d1148435fa2cadb36da7df2bd865a3d/EIPS/eip-649.md",
+	author = "Afri Schoedon and Vitalik Buterin",
+	title = "EIP-649: Metropolis Difficulty Bomb Delay and Block Reward Reduction",
+	year = "2017",
+	month = "June",
+}
+
+@Misc{EIP-2,
+	url = "https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2.md",
```
pirapira: Why do other URLs contain `/blob/<hash>` while this one has `/blob/master`?

jamesray1: Changed. It did cross my mind at the time but I didn't think more about it.

Paper.tex
```diff
@@ -482,6 +482,8 @@ \subsubsection{Block Header Validity}
 H'_i &\equiv \max(H_i - 3000000, 0)
 \end{align}
 
+Note that $\mindifficulty$ is the difficulty of the genesis block. The \textit{Homestead} difficulty parameter, $\homesteadmod$, is used to affect a dynamic homeostasis of time between blocks, as the time between blocks varies, as discussed below, as implemented in EIP-2 \cite{EIP-2}. In the Homestead release, $\expdiffsymb$ results in the difficulty increasing every 10000 blocks, which puts upward pressure on the time between blocks, causing it to slowly accelerate, and putting time pressure on transitioning to proof-of-stake. This effect, known as the "difficulty bomb", or "ice age", was explained in EIP-649 \cite{EIP-649} and delayed and implemented earlier in EIP-2 \cite{EIP-2}. $\homesteadmod$ was also modified in EIP-100 with the use of $x$, the adjustment factor above, and the demoninator 9, in order to target the mean block time including uncle blocks \cite{EIP-100}.
```
pirapira: Does the exponential factor "slowly accelerate"?  I thought the opposite.

jamesray1:
I am quoting from EIP 649.

jamesray1:
Specifically, the original PR, rather than the final merged EIP.
https://github.com/ethereum/EIPs/pull/669/files

jamesray1:
It's also mentioned in the same place in the summary of the final EIP.
https://github.com/ethereum/EIPs/blob/master/EIPS/eip-649.md

pirapira:
@jamesray1 will you try to fix them?

pirapira:
Just because EIPs use misleading words, YP doesn't need to repeat them.

jamesray1:
The difficulty bomb is what is described as slowly accelerating with average block times increasing.

The average block times are increasing due to the difficulty bomb (also known as the "ice age") slowly accelerating.
https://github.com/ethereum/EIPs/blob/master/EIPS/eip-649.md#simple-summary

pirapira:
@jamesray1 that reads like the bomb is accelerating, whatever that means.  On the other hand, this PR reads like the time between blocks accelerates, which is plain wrong.

jamesray1:
If the EIP is misleading then it should be edited. But is it misleading?

jamesray1:
`+ int(2**((block.number // 100000) - 2))`
That's the difficulty bomb, or "exponential difficulty adjustment component", from EIP 2.

jamesray1:
That is added to the block time difference, which would cause the time difference to slowly accelerate as the block numbers increase.

Paper.tex
pirapira: That sounds good.

Paper.tex
```diff
@@ -482,6 +482,8 @@ \subsubsection{Block Header Validity}
...
...omesteadmod$, is used to affect a dynamic homeostasis of time between blocks, as the time between blocks varies, as discussed below, as implemented in EIP-2 \cite{EIP-2}. In the Homestead release, the exponential difficulty symbol, $\expdiffsymb$ causes the difficulty to slowly increase (every 100,000 blocks) at an exponential rate, and thus increasing the block time difference, and putting time pressure on transitioning to proof-of-stake. This effect, known as the "difficulty bomb", or "ice age", was explained in EIP-649 \cite{EIP-649} and delayed and implemented earlier in EIP-2 \cite{EIP-2}. $\homesteadmod$ was also modified in EIP-100 with the use of $x$, the adjustment factor above, and the demoninator 9, in order to target the mean block time including uncle blocks \cite{EIP-100}. Finally, in the Byzantium release, with EIP-649, the ice age was delayed by creating a fake block number, $H'_i$, which is obtained by substracting three million from the actual block number, which in other words reduced $\expdiffsymb$ and the time difference between blocks, in order to allow more time to develop proof-of-stake and preventing the network from "freezing" up due to network congestion.
```
pirapira: "due to network congestion" sounds inaccurate.

After certain time, even if the transactions are very few, the network freezes up.

jamesray1:
Revised.

Issue #366, opened 2017-10-01
ε-TeX needed, using PDF-TeX, use pdf(ε)-TeX?


jamesray1:
The csquotes package says that ε-TeX, however, I believe we are using PDF-TeX. 
https://ctan.org/tex-archive/macros/latex/contrib/csquotes?lang=en

I had a read [here](http://latex.org/know-how/latex/55-latex-general/475-e-tex) and found the figure 1 diagram:
http://latex.org/images/tex-engines.png 
Seems like we could use pdf(ε)-TeX?

jamesray1:
Related to PR #365

jamesray1:
Alternatively add [texlive](https://ctan.org/pkg/texlive).

jamesray1:
I just saw in the Travis build that texlive is added.

jamesray1:
According to [here](https://tex.stackexchange.com/a/13070/143781) it would be good to try using [LuaTeX](http://luatex.org/) or XeTeX. I figure that LuaTeX is better to try first because it is a child of pdf(ε)-TeX (which in turn is a child of PDFTeX).

jamesray1:
LuaTex and XeTeX are part of [texlive](https://www.ctan.org/pkg/texlive).

jamesray1:
This is what I tried unsuccessfully to get Unicode characters to render properly:
```
%You don't use inputenc with XeTeX.
%\usepackage[utf8]{inputenc}
%\DeclareUnicodeCharacter{00A0}{ }
%\DeclareUnicodeCharacter{C3 ab}{"e}
%\DeclareUnicodeCharacter{C3 bc}{"u}
%For testing: \DeclareUnicodeCharacter{AD}{\Huge [ICI]}
```

PR #437, opened 2018-01-20
Creating dependency files for the csquotes package, e.g. to use opening double quotes

jamesray1:
Related: https://github.com/ethereum/yellowpaper/issues/366. Required for any changes that use an opening double quote.

### DIFF
--- a/csq-compat.def
+++ b/csq-compat.def
@@ -1,0 +1,103 @@
+\ProvidesFile{csq-compat.def}
+  [2017/03/11 v5.2b csquotes backwards compatibility (JAW)]
+
+\def\csq@getversion#1{%
+  \expandafter\csq@getversion@i
+  \expandafter{\csq@vertonum#1\relax}}
+
+\def\csq@getversion@i#1{%
+  \ifcsdef{csq@version@#1}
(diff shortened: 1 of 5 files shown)


jamesray1:
They are needed to use features related to quotations, such as using opening double quotes instead of two single quotes. 
I found them by searching for them when it mentioned them in but the failed Travis build.
https://ctan.org/tex-archive/macros/latex/contrib/csquotes
https://ctan.org/pkg/etoolbox?lang=en

pirapira:
@jamesray1 even the plain `tex` can interpret `` (backquote repeated twice) as the opening-double-quote.

pirapira:
The Ubuntu package `texlive-latex-extra` should also contain `csquotes.sty`.

jamesray1:
Replaced by #444.

jamesray1:
> @jamesray1 even the plain tex can interpret `` (backquote repeated twice) as the opening-double-quote.

Yoichi, I understand that, but some people may edit the document without being familiar with the LaTeX rule to use two single backticks as a double quote, and just use a double quote as an opening quote (which quite frankly, I find is an odd rule). So adding the csquotes package helps as a fix for that.

PR #438, opened 2018-01-22
Gas usage of CALL should first take C_CALLGAS into account

pirapira:
Before this change, when the CALL instruction got gas refuned from the callee, the caller didn't losing the initial gas passed to the callee.
This change makes the caller to lose the initial gas passed to the callee before receiving the remainder.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,18 @@
+sudo: required
+before_install:
+- sudo apt-get -qq update
+- sudo apt-get install texlive texlive-latex3
+script:
+- ./build.sh
+deploy:
+  provider: script
+  script: ./travis_deploy.sh
+  skip_cleanup: true
+  on:
+    branch: master
+env:
+  global:
+    - ENCRYPTION_LABEL="19a81de38b62"
+    - COMMIT_AUTHOR_EMAIL="chris@ethereum.org"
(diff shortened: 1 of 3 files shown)


pirapira:
This is a reincarnation of #304 (the first commit).  The second commit of #304 goes to `tangerine-whistle`.

.travis.yml
```diff
@@ -0,0 +1,18 @@
...
+- sudo apt-get -qq update
+- sudo apt-get install texlive texlive-latex3
+script:
+- ./build.sh
+deploy:
+  provider: script
+  script: ./travis_deploy.sh
+  skip_cleanup: true
+  on:
+    branch: master
```
nicksavers: If you merge this into `frontier`, the branch for the script should also use that branch.

pirapira: Really?  I don't want to deploy PDFs from `frontier`.

pirapira:
Since I have
```
on: master
```

The PDF will not be deployed when you merge this into `frontier`.  See the Travis document on this:
https://docs.travis-ci.com/user/deployment#Conditional-Releases-with-on%3A

pirapira:
I prefer to have identical `.travis.yml` on all branches.

pirapira:
@nicksavers can we move ahead here?

PR #442, opened 2018-01-22
The all-but-one-64th calculation should occur after the memory expansion fee is taken

pirapira:
[EIP 150](https://github.com/ethereum/EIPs/issues/150) introduced the all-but-one-64th computation.  The all-but-one-64th computation should be done after the memory expansion fee is taken out.


pirapira:
This was originally the second commit of #304.

PR #443, opened 2018-01-22
Make sure that divisions and modulo operations are not subject to mod 2^{256}

pirapira:
At the beginning of the table of instructions, it's written that all operations are subject to modulo 256-bit, so when the modulo is not applicable, the exception should be mentioned.


jamesray1 (approved):

Mitumaru (approved):

pirapira:
I think this has gotten too stale.

PR #446, opened 2018-01-22
"where" following an equation does not introduce a new paragprah

pirapira:
Or, even a new sentence.


pirapira:
ahh, this one should target `frontier`!

PR #447, opened 2018-01-22
Grammar fixes: add "a", omit 'the"

jamesray1:


README.md
```diff
@@ -5,7 +5,7 @@
 
 The Yellow Paper is a formal definition of the Ethereum protocol, originally by Gavin Wood, currently maintained by Nick Savers and with contributions from many people around the world.
 
-It is free culture work, licensed under the Creative Commons Attribution Share-Alike (CC-BY-SA) version 4.0.
+It is a free culture work, licensed under Creative Commons Attribution Share-Alike (CC-BY-SA) version 4.0.
```
pirapira: "work" is usually not counted.

jamesray1: It depends. In this case, I think it should be, as it is a creative work, which is countable. The verb "work" is not countable.

pirapira: OK.

PR #421, opened 2018-01-19
Edit details for Gavin Wood author header

jamesray1:
@gavofyork should approve this before merging.

PR #449, opened 2018-01-22
Change Ethcore reference to Parity in YP

5chdn:
cc @gavofyork


jamesray1:
@5chdn, this is a duplicate of #421. I suggest that if you want to change this then you email i@gavwood.com to ask Gavin directly, as I suspect that he may be ignoring Github notifications for this repo.

pirapira:
@jamesray1 the diff is similar to yours but it comes from somebody working with Parity.

pirapira:
@nicksavers what do you think?

PR #454, opened 2018-01-23
Unix's [sic] time

jamesray1:


pirapira:
The phrase does not about "Unix time" but one of the Unix's interface functions.

PR #458, opened 2018-01-23
hypertargets lines 100–199

jamesray1:


pirapira:
Why does Travis fail?

jamesray1:
I don't know. The error in the build is odd.

```
! Missing $ inserted.
<inserted text> 
                $
l.104 ...ate_transition}{Upsilon_state_transition}
```
\hypertarget{Upsilon_state_transition}{Upsilon_state_transition} isn't in math mode.

jamesray1:
Now the error has changed:
```
! LaTeX Error: Something's wrong--perhaps a missing \item.
See the LaTeX manual or LaTeX Companion for explanation.
Type  H <return>  for immediate help.
 ...                                              
                                                  
l.199 \item[nonce]
                   A scalar value equal to the number of transactions sent f...
```

jamesray1:
With the new `\nop` command created in the preamble, after this PR is merged I'll be able to similarly replace \hypertarget{foo}{} with \nop{foo} in other PRs, and move it to the same line to direct to, which it will do.

pirapira:
@jamesray1 somehow the name `nop` means very little.  Can you use a more informative name?

jamesray1:
OK let's decide before I make the change. How about furl? It's short for fixed URL (and conveniently pronounceable as one word).

pirapira:
It's a target, or a marker, or a destination, or https://github.com/ethereum/yellowpaper/pull/458#issuecomment-360785506 .

jamesray1:
OK OK I didn't think that through. What about dest, or targ, or mark?

pirapira:
linkdest?

jamesray1:
OK let's use that.

PR #461, opened 2018-01-23
hypertargets lines 100–670

jamesray1:


jamesray1:
Split into PRs #458 to #465 other than this one.

PR #472, opened 2018-01-23
changed ordering of receipt items to correct client implementation

### LINKED ISSUE #471: Receipt Spec Incorrect
Although the paper says:
image
The stated order of items being: 
`(R_u, R_b, R_\mathbf{l}, R_{s'})`

However, I observe the ordering on mainnet (which must be how clients are implementing it) to be:
`( R_{s'}, R_u, R_b, R_\mathbf{l})`

Obviously the simple thing to do is to correct the paper to match whatever geth/parity are already doing.

zmitton:
Fixes #471


pirapira:
I restarted the Travis build (whose failure seems like an irrelevant accident).

PR #482, opened 2018-01-25
\hypertarget{R l}{}

jamesray1:


jamesray1:
I think if you build this it will still have an empty line, IIRC from testing this.

pirapira:
The indentation before `Thus` has disappeared.  So this PR joins two paragraphs into one.

PR #493, opened 2018-01-25
hypertargets for Ethash parameters

jamesray1:


jamesray1:
I have put every hypertarget before the line that it refers to, because when you are directed to the hypertarget, it actually displays the next line at the top of the screen. So to actually include the line you want to display, you need to have the target on the preceding line.

pirapira:
That way, mistakes can happen when people insert lines.
```
A \target{C}\\
C
```
into
```
A \target{C} \\
B \\
C
```
Is there a good workaround?  Do all hyperref users tolerate this?

jamesray1:
I don't know. [I have asked around](https://tex.stackexchange.com/questions/411987/how-to-fix-hypertargets-directing-links-to-the-line-after-the-target).

pirapira:
Similar questions with answers

https://tex.stackexchange.com/questions/17057/hypertarget-seems-to-aim-a-line-too-low

https://tex.stackexchange.com/questions/2068/hyperlinks-to-a-bibliography-are-one-line-off

jamesray1:
I will need to change all the hypertargets. Can we just merge as is? Hopefully people will move the hypertarget if they add lines.

pirapira:
The future communication costs about line arrangement outweighs the current fix.

PR #494, opened 2018-01-25
Add commands to manage URL breaks + glossaries package

jamesray1:


jamesray1:
Related:

https://tex.stackexchange.com/questions/60299/a-doesnt-match-its-definition-problem-that-i-dont-understand

PR #496, opened 2018-01-25
Find and replace all _([:alpha][:alnum]*) with _mathrm{$1} ...

jamesray1:
... using Libre Open Office with regular expressions enabled as an option. Man it'd be really good if you could change a commit summary after committing, on Github. I passed on that feedback to support@github.com. Additionally, there are now no search results for _{([:alpha][:alnum]*)} (also with regular expressions ticked with find and replace on LOO).

### DIFF
--- a/Paper.tex
+++ b/Paper.tex
@@ -256,7 +256,7 @@ A transaction (formally, $T$) is a single cryptographically-signed instruction c
 
 \begin{description}
 \item[nonce] A scalar value equal to the number of transactions sent by the sender; formally $T_n$.
-\item[gasPrice] A scalar value equal to the number of Wei to be paid per unit of \textit{gas} for all computation costs incurred as a result of the execution of this transaction; formally $T_p$.
+\item[gasPrice] A scalar value equal to the number of Wei to be paid per unit of \textit{gas} for all computation costs incurred as a result of the execution of this transaction; formally $T_mathrm{p}$.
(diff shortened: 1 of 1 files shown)


Paper.tex
```diff
@@ -281,15 +281,15 @@ \subsection{The Transaction} \label{ch:transaction}
 
 \begin{equation}
 L_T(T) \equiv \begin{cases}
-(T_n, T_p, T_g, T_t, T_v, T_\mathbf{i}, T_w, T_r, T_s) & \text{if} \; T_t = \varnothing\\
-(T_n, T_p, T_g, T_t, T_v, T_\mathbf{d}, T_w, T_r, T_s) & \text{otherwise}
+(T_n, T_mathrm{p}, T_g, T_t, T_v, T_\mathbf{i}, T_w, T_r, T_s) & \text{if} \; T_t = \varnothing\\
+(T_n, T_mathrm{p}, T_g, T_t, T_v, T_\mathbf{d}, T_w, T_r, T_s) & \text{otherwise}
```
pirapira: Other elements of the tuple should be treated the same way.

jamesray1: Hmm, I wonder why that wasn't picked up.

jamesray1: Ah I know what I did wrong.

jamesray1:
Replaced by #499.

PR #497, opened 2018-01-25
Add a line break before a tx address so that it doesn't run over the …

jamesray1:
…margin.


jamesray1:
Yes, it does, and for good reason. The address runs over the margin and edge of the page at the moment.

pirapira:
Well, you'll need to move the `\\` every time the layout changes.

Please try https://ctan.org/pkg/seqsplit .

jamesray1:
That works!

jamesray1:
I did some more testing, and found that if the hypertarget is in a text-mode location, links to it work normally (they show the target line at the top of the page). However, if the hypertarget is in math mode (of any kind: in-line or in display-math or in equation arrays), then it will direct to the line after the target.

PR #499, opened 2018-01-26
Find and replace all _([:alpha:][:alnum:]*) with _mathrm{$1}

jamesray1:
Using Libre Open Office, with regular expressions ticked as an option.


pirapira:
More crucially, the backslash is missing before `mathrm`.

jamesray1:
> I guess you don't need to do mathrm in URLs 

Obviously that was an oversight.

jamesray1:
#515

PR #504, opened 2018-01-26
Blocks function as a ledger

jamesray1:


nicksavers:
The aggregate of transactions in blocks are the journal. The world state is the ledger.

PR #512, opened 2018-01-26
Add missing introductory text for $\Upsilon^s$.

benjaminion:
For consistency: there are now three super-scripted versions of $\Upsilon$.


benjaminion:
There should probably have been a rebase after #511 before this was merged, so it's now a little inconsistent.  Let me submit another PR to tidy up.

PR #514, opened 2018-01-29
Tidy up change to z for status code.

benjaminion:
A combination of PRs #511 and #512 being committed led to some inconsistency. This fixes it and tidies up a few superfluous curly braces at the same time which I previously missed.


jamesray1:
Actually AIUI it's better to use _{foo} rather than _foo, to be clear what it is you intend to be subscripted.

pirapira:
@jamesray1 when I see Knuth's TeX book https://ctan.org/pkg/texbook?lang=en he uses `_i` and so, for one-character subscripts, but I agree that the TeX source is sometimes confusing.  I just merge whenever I don't see a mistake; when in doubt, I typeset.

PR #520, opened 2018-01-31
Remove \small and \tiny in math mode because they have no effects

pirapira:
and caused error messages like
```
LaTeX Font Warning: Command \tiny invalid in math mode on input line 623.
```


jamesray1:
I guess that this should be merged into ethereum: master, as well, right?

pirapira:
I'll merge `frontier` into `homestead`, and remove warnings, and so on.

jamesray1:
OK.

Issue #517, opened 2018-01-31
Inconsistency between yellow paper and Ethash wiki


pgebal:
Hi,
Please take a look at eq. 230 in yellow paper and [Ethash wiki](https://github.com/ethereum/wiki/wiki/Ethash) implementation of get_cache_size function. If the function was implemented as in yellow paper we would have `sz -= HASH_BYTES` instead of `sz -= 2 * HASH_BYTES`. ethereumj and pyethereum implementations follow Ethash wiki. Which version is correct? Should I follow Ethash wiki or yellow paper when implementing Ethash?

pirapira:
I'll compare both with implementations.

pirapira:
Wait, the formula in YP doesn't even terminate.

pirapira:
@pgebal does this look good? https://github.com/ethereum/yellowpaper/pull/519

pirapira:
@pgebal I realized something.  Whichever definition you choose, they result in the same cache size, because no even number is prime.

pirapira:
Whenever (sz / HASH_BYTES) is an odd number at least three, the loop
```
    while not isprime(sz / HASH_BYTES):
        sz -= HASH_BYTES
```
runs even times. Say `119-> 118 (even, not a prime) -> 117 -> 116 (even, not a prime) -> 115`.

So, the loop can be optimized, merging every two steps, into
```
    while not isprime(sz / HASH_BYTES):
        sz -= 2 * HASH_BYTES
```

pgebal:
You are right. Thank you.

pirapira:
I keep my pull requests open.  Different formulations are confusing.  Thank you for pointing them out.

PR #521, opened 2018-01-31
Fix the cache size computation in ethash (frontier)

pirapira:
See https://github.com/ethereum/yellowpaper/issues/517


nicksavers:
https://github.com/ethereum/go-ethereum/blob/master/consensus/ethash/algorithm_go1.8.go#L39

PR #535, opened 2018-02-05
hyperlinks for state transition functions

jamesray1:


Paper.tex
```diff
@@ -319,7 +319,7 @@ \subsection{The Transaction} \label{ch:transaction}
 \mathbb{B}_{0} & \text{otherwise}\end{cases}
 \end{equation}
 
-\subsection{The Block} \label{ch:block}
```
pirapira: This label is used twice, so `\ref{ch:block}` occurrences need to be replaced as well.

PR #536, opened 2018-02-05
\hyperref{https://medium.com/@VitalikButerin/the-meaning-of-decentralization-a0c92b76a274}{decentralised}

jamesray1:
This probably doesn't need to be a citation/reference, as it's just a blog post.


pirapira:
Ahm, I wouldn't do this.  It's a too centralized interpretation of decentralization.

jamesray1:
Haha, hmm, fair enough. Maybe could add some other references for diversity.

PR #540, opened 2018-02-05
\linkdest{ell}{} (add {})

jamesray1:


jamesray1:
Actually I just tested that you don't need the second {}.
```
\begin{abstract}
The\linkdest{blockchain_paradigm} blockchain paradigm when coupled with cryptographically-secured transactions has demonstrated its utility through a number of projects, with Bitcoin being one of the most notable ones. Each such project can be seen as a simple application on a decentralised, but singleton, compute resource. We can call this paradigm a transactional singleton machine with shared-state.

\hyperlink{foo}{Eth}ereum implements this \hyperlink{blockchain_paradigm}{paradigm} in a generalised manner. Furthermore it provides a plurality of such resources, each with a distinct state and operating code but able to interact through a message-passing framework with others. We discuss its design, implementation issues, the opportunities it provides and the future hurdles we envisage.

$\linkdest{foo}foo$
\end{abstract}
```

jamesray1:
@pirapira, FYI.

Issue #541, opened 2018-02-05
Fix errors in the build output


jamesray1:
E.g.:
Package natbib Warning: Citation `buterin2013ethereum' on page 1 undefined on i
nput line 99.
Package natbib Warning: Citation `dagger' on page 14 undefined on input line 12
07.
LaTeX Warning: Reference `ch:Terminology' on page 2 undefined on input line 114
.
Underfull \hbox (badness 10000) in paragraph at lines 204--205
[]\OT1/cmr/m/n/9 Having in-tro-duced the ba-sic con-cepts be-hind

jamesray1:
buterin2013ethereum does exist in Biblio.bib.

jamesray1:
No, as above with #542, this doesn't help, the build still says `No file Paper.aux.` and `No file Paper.bbl.`. You still get the errors: `Package natbib Warning: Citation `buterin2013ethereum' on page 1 undefined on i
nput line 99.`

PR #542, opened 2018-02-05
Test removing Paper.aux Paper.bbl

jamesray1:
Related: #541.


jamesray1:
Well, the build still says `No file Paper.aux.` and `No file Paper.bbl.`.

PR #549, opened 2018-02-05
Update and rename build.sh to b.sh

jamesray1:
b.sh is short for build.sh. Having a very short file name saves time typing this command many times into the terminal, e.g. when making a small change, then building, and repeating these two steps many times.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ before_install:
 - sudo apt-get install texlive texlive-latex3
 - sudo apt install texlive-latex-extra
 script:
-- ./build.sh
+- ./b.sh
 deploy:
   provider: script
   script: ./travis_deploy.sh
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ The paper also comes as a single ``latex`` file ``Paper.tex``, which is built as
 ```
 git clone https://github.com/ethereum/yellowpaper.git
 cd yellowpaper
-./build.sh
+./b.sh
 ```
 This will create a PDF version of the Yellow Paper. Following building, you can also use standard `pdflatex` tools like http://latex.informatik.uni-halle.de/latex-online/latex.php for compiling/preview.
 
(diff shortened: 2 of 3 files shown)


pirapira:
Even if a tab-completion doesn't work for you, I'd prefer keeping the filename descriptive.

jamesray1:
I'll be making changes gradually.

fulldecent:
Usually I keep the shell open and just press <kbd>up</kbd>, <kbd>enter</kbd>

fulldecent:
Also... just in case you haven't see it, https://github.com/robbyrussell/oh-my-zsh

jamesray1:
@fulldecent, I had a quick look at https://github.com/robbyrussell/oh-my-zsh, but I'm not sure how that helps me for this. Using up and enter may help if ./build.sh was entered recently.

fulldecent:
I use a *separate* terminal for this. Then the up-enter works because I have nothing else ever going on in there.

oh-my-zsh is just a note because you said you're using bash. I'm a long-time bash user too, my life got a lot better when I found omz

jamesray1:
Ah OK, makes sense. For me, I often create separate folders for each test (using `git checkout -b` doesn't actually change the files on Nautilus), although I rarely go back to a folder once I built it, but still, rarely is not never. Using a gnote to copy and paste ./build.sh isn't really any faster than just typing ./build.sh. But in any case it's not that big a deal to type ./build.sh, it just gets rather tedious when you are doing it frequently, although that is often with one folder, so having one terminal just for ./build.sh for each folder would be handy.

jamesray1:
Nice trick with the keyboard glyphs `<kbd>up</kbd` = <kbd>up</kbd

jamesray1:
Powerline fonts seemed to install correctly, but:
```
james@james-Streacom:~$ echo "\ue0b0 \u00b1 \ue0a0 \u27a6 \u2718 \u26a1 \u2699"
\ue0b0 \u00b1 \ue0a0 \u27a6 \u2718 \u26a1 \u2699
```
I opened an issue.
https://github.com/powerline/fonts/issues/274

PR #550, opened 2018-02-07
Cleanup ECREC definition and a few other things

ldct:
I was searching for the gas cost for the precompile invoked in solidity as `ecrecover` and couldn't find it easily. Turns out it's called ECREC in the yellow paper. This PR

- Adds `ecrecover` to make it more searchable
- Applies consistent abbreviation in that section
- Puts a forward reference to the ECDSARECOVER function, since it is defined in appendix F after it is used in appendix E
- Removes an empty equation from the SHA256/RIP160 definitions

### DIFF
--- a/Paper.tex
+++ b/Paper.tex
@@ -26,7 +26,7 @@
 
 %This should be the last package before \input{Version.tex}
 \PassOptionsToPackage{hyphens}{url}\usepackage{hyperref}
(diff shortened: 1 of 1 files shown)


ldct:
any reviewers had a chance to look at this? @pirapira ?

pirapira:
It's 8th on my list now.

Paper.tex
```diff
@@ -1148,6 +1148,7 @@ \subsection{State \& Nonce Validation}\label{sec:statenoncevalidation}
 B'_{\mathrm{m}} & = & m \quad \text{with } (x, m) = \mathtt{PoW}(B^*_{\hcancel{n}}, n, \mathbf{d}) \\
 B^* & \equiv & B \quad \text{except:} \quad B^*_{\mathrm{r}} = r(\Pi(\Gamma(B), B))
 \end{eqnarray}
+
```
pirapira: I think it was better without this empty line.  Now the next line lives in an independent paragraph.

PR #561, opened 2018-02-08
The up-front cost hyperlinks

jamesray1:
It's pretty simple anyway but nevertheless...


jamesray1:
Or not terse enough, perchance?

PR #579, opened 2018-02-09
Add parity bit explanation for EIP 155 signatures

Sword-Smith:
T_w describe the parity of the y value of the curve point of the signature. This is true for both the original (pre-EIP 155) signature scheme and for the EIP 155 signature scheme.


Paper.tex
```diff
@@ -1659,7 +1659,7 @@ \section{Signing Transactions}\label{app:signing}
...
+Where $p_{\mathrm{u}}$ is the public key, assumed to be a byte array of size 64 (formed from the concatenation of two positive integers each $< 2^{256}$) and $p_{\mathrm{r}}$ is the private key, a byte array of size 32 (or a single positive integer in the \hypertarget{v}{}aforementioned range). It is assumed that $v$ is either the `recovery identifier' or `chain identifier doubled plus 35 or 36'. The recovery identifier is a 1 byte value specifying the parity and finiteness of the coordinates of the curve point for which $r$ is the x-value; this value is in the range of $[27, 30]$, however we declare the upper two possibilities, representing infinite values, invalid. The values 27 and 35 represent an even $y$ value and 28 and 36 represent an odd $y$ value.
```
pirapira: Now it's becoming vague what `27` `35` `28` and `36` are.  Two of them are recovery identifiers, and the other two are numbers appearing in the computation of $v$.

Maybe it's cleaner to say, an odd $v$ value represents an even $y$ value?

Sword-Smith: I will take a look at rewriting this section. I think it is possible to formulate it better.

pirapira: I'm waiting for your rewrite.

Sword-Smith: I have created a new pull request in #679. I suggest we close this and look at #679 instead.

Sword-Smith:
I have created a new pull request in #679. I suggest we close this and look at #679 instead.

pirapira:
Thank you!

PR #581, opened 2018-02-12
Use build/ directory for building

fulldecent:
This uses a separate directory for building. This makes development cleaner because you can get a clean state by removing the `build/` directory.


axic:
Need to update `.gitignore` if the files are moved.

fulldecent:
The `.gitignore` does not use absolute paths for file names, so this is unnecessary.

Reference: https://github.com/ethereum/yellowpaper/blob/master/.gitignore

PR #583, opened 2018-02-16
Fix subsequent build runs

fulldecent:
Sorry, I broke the build with https://github.com/ethereum/yellowpaper/pull/581

Here is the fix.

The problem is that build works great. But it fails the *second* time you run it.

We're good now, my bad.


nicksavers:
Should be fixed with #584

fulldecent:
Yes. Closing.

PR #585, opened 2018-02-16
Clearer OOG exception wording, fixes #84

### LINKED ISSUE #84: Typo in exceptions regarding state changes
"Like the out-of-gas (OOG) exception, they not leave state changes intact."

Shouldn't it be ?
"Like the out-of-gas (OOG) exception, they leave the state intact"

fulldecent:
From @SergioDemianLerner


nicksavers:
I agree that it could be made clearer. Although neither is really correct. The state is not left intact. A failed transaction is still added to the state.

PR #603, opened 2018-02-16
Mist

jamesray1:


nicksavers:
This seems to undo some of your changes and needs a rebase.

jamesray1:
Rebases are not as intuitive to do as to just make another PR, so I'll do that instead.

jamesray1:
#618

PR #617, opened 2018-02-18
Fix-links-block-nonce-account-nonce-Gtxcreate-etc

jamesray1:


Paper.tex
```diff
@@ -321,7 +321,7 @@ \subsection{The Transaction} \label{subsec:transaction}
 \mathbb{B}_{0} & \text{otherwise}\end{cases}
 \end{equation}
 
-\subsection{The Block} \linkdest{block}{}
+\subsection{The Block}\linkdest{block}\label{subsec:The_Block}
```
pirapira: How is this parsed?  The second argument of `\linkdest` is `\label`?

jamesray1: Not quite. `linkdest` will just refer to the line that it's on. But when you have `\hyperlink{block}{\ref{subsec:The_Block}}`, it will display:
screenshot from 2018-02-20 13-12-22
Try building it!

jamesray1: Then the link will jump to:
screenshot from 2018-02-20 13-13-40

pirapira: I built it before writing my comment, of course.  The build was a success, but LaTeX sometimes does nonsense.

pirapira: Ah, `\linkdest` takes only one argument.  I missed that.

PR #618, opened 2018-02-19
@misc{Mist,

jamesray1:


pirapira:
I don't know if this should be in YP.  Mist is just one browser.

jamesray1:
But it is the only Ethereum dapp browser at the moment that I am aware of. It is certainly helpful to give an example of an actual implementation, rather than just giving a vague idea.

nicksavers:
Mist might be the only actual Ethereum Browser, but the text should not be about a specific implementation. It's about the concept.

PR #619, opened 2018-02-19
Make sure that Travis fails when latex fails

pirapira:
`set -e` seems to be non-effective in Travis build


pirapira:
Before this change, Travis didn't fail even with a broken LaTeX file.

https://travis-ci.org/ethereum/yellowpaper/builds/342650328?utm_source=github_status&utm_medium=notification

PR #626, opened 2018-02-20
Replace all \$([ceg-hmnz])\$ with \$\mathrm{$1}\$ using Libre Open Office with match case and regular expressions ticked as options.

jamesray1:


Paper.tex
```diff
@@ -181,7 +181,7 @@ \section{Conventions}\label{ch:conventions}
 
 Tuples are typically denoted with an upper-case letter, \eg $T$, is used to denote an Ethereum transaction. This symbol may, if accordingly defined, be subscripted to refer to an individual component, \eg \hyperlink{transaction_nonce}{$T_{\mathrm{n}}$}, denotes the nonce of said transaction. The form of the subscript is used to denote its type; \eg uppercase subscripts refer to tuples with subscriptable components.
 
-Scalars and fixed-size byte sequences (or, synonymously, arrays) are denoted with a normal lower-case letter, \eg $n$ is used in the document to denote a \hyperlink{transaction_nonce}{transaction nonce}. Those with a particularly special meaning may be Greek, \eg $\delta$, the number of items required on the stack for a given operation.
+Scalars and fixed-size byte sequences (or, synonymously, arrays) are denoted with a normal lower-case letter, \eg $\mathrm{n}$ is used in the document to denote a \hyperlink{transaction_nonce}{transaction nonce}. Those with a particularly special meaning may be Greek, \eg $\delta$, the number of items required on the stack for a given operation.
```
pirapira: I think this edit is mistaken about the `normal lower-case letter`.  Scalars and fixed-size byte sequences should be typeset with italic.

jamesray1: Hmm, OK, I guess I changed it like this (overlooking this part) because n, etc, are also used as a field, e.g. of transactions, accounts, blocks, where they are typeset in roman font. So it seems like a bit of a contradiction that they are typeset differently.

pirapira: Let's keep it inconsistent for now, and tidy up the document first.

jamesray1: So what, just leave this PR open and make a decision whether to merge or close it some time later? Why not make a decision now? I guess we want to avoid merge conflicts.

pirapira: If you want a decision now, here
* scalar or fixed-length without subscripts `$x$`
* scalar or fixed-length component of a record `$T_{\mathrm{n}}$`
* scalar of fixed-length component of an array specified by an index `$R_i$`
* variable-length $\mathbb{x}$
* variable-length component of a record $\mu_{\mathbb s}$

jamesray1: Interesting conventions, is there a style guide for that?

pirapira: No, just observing what's happening now.

Mitumaru (approved):

PR #629, opened 2018-02-20
(\cite{ASICSmithMJS}) (\cite{Courtois2014})

jamesray1:


Biblio.bib
```diff
@@ -1,3 +1,13 @@
+@book{ASICSmithMJS,
+	title = {{Application-Specific Integrated Circuits}},
+	author = "Smith, Michael John Sebastian",
+	publisher = "Addison-Wesley",
+	year = "1997",
+	ISBN = "0201500221",
+	note = "A preview is also available at \url{https://www.amazon.com/Application-Specific-Integrated-Circuits-Michael-Smith/dp/0321602757}, with the archive:
```
pirapira: I don't think YP should lead traffic to Amazon.

pirapira: Traditionally when you reference a book, you don't need to show how to see or buy it.

PR #637, opened 2018-02-22
Fix the font of \alpha_w and \delta_w

pirapira:
Since w is a variable for an opcode, it should be typeset in italic.


pirapira:
misclicked.

PR #643, opened 2018-02-23
Use well-known \mathbb{N} for natural numbers

fulldecent:
Currently `\mathbb{P}` is in the text used to define "positive integers".

This concept is much better known as a "natural numbers" with the symbol `$\mathbb{N}$`. This PR does not change the name "positive integers", but it does adopt the well known symbol.


fulldecent:
Also, I notice there are several places where zero is used, in violation of the "positive integer" definition.

One example: H_i = 0.

But that is outside the scope of this PR.

DanielMReed:
Thank you! I mentioned this in the gitter, was meant to add a PR

PR #645, opened 2018-02-23
Typo

fulldecent:


nicksavers:
The size of **_a_** single stack item is still 256-bit. But this works too.

PR #647, opened 2018-02-23
Remove verifiable computation

fulldecent:
There was a note about verifiable computing:

> Finally, verifiable computation, if made generally available and efficient enough, may provide a route to allow the proof-of-work to be the verification of final state.

Just to be clear, verifiable computing would completely obviate the need for Ethereum. There is no need to consider other minutiae of how verifiable computing would would impact the then-dead Ethereum project.


pirapira:
It's not obvious to me why practical verifiable computing would kill Ethereum. How does one prevent double spending just with verifiable computing?

fulldecent:
Because you can use a single host and make a synchronous call that goes through or fails.

Ethereum double-spending prevention is only necessary because you have to wait for the transaction to be mined which is asynchronous and may execute out-of-order with another spend.

pirapira:
In that case, users rely on this single host for availability.  Ethereum at least tries to make sure that a single user can keep the network alive.

fulldecent:
I think there's a better argument here, but I don't have it now, closing issue. Thanks for entertaining the thought.

PR #660, opened 2018-03-04
Fix Elliptic Curve group addition equations

ValarDragon:
Currently the Yellow Paper's specification for the elliptic curve group law has that the definition for (P + P) for a point P in the elliptic curve is (0,0) (The representation for the point at infinity). This adds the correct equation for elliptic curve addition in short Weierstrass form for addition when P = Q in the expression P + Q.


pirapira:
Thank you.  That sounds problematic.

PR #661, opened 2018-03-05
For more mathematical symbols, see \href{https://en.wikipedia.org/wik…

jamesray1:
…i/List_of_mathematical_symbols}{here}.

### DIFF
--- a/Paper.tex
+++ b/Paper.tex
@@ -2562,4 +2562,6 @@ Symbol & Latex Command & Description \\
 \linkdest{bigvee}$\bigvee$ & \verb|\bigvee| & This is the least upper bound, supremum, or join of all elements operated on. Thus it is the greatest element of such elements (\cite{Davey2002_zbMATH01748069}).\\
 \bottomrule
 \end{tabu*}
+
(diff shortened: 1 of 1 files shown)


pirapira:
I think we should try to explain the symbols ourselves.

DanielMReed:
Agree with @pirapira , if someone has a list of symboks that they feel needs explaining I can create a relevant entry at the end of the YP, shall I create an issue for people to submit them?

pirapira:
@DanielMReed feel free to.

jamesray1:
Why re-invent the wheel?

https://en.wikipedia.org/wiki/List_of_mathematical_symbols is comprehensive, why spend the extra labour on this when it is already done? There is much work to do on the development of Ethereum, it's better not to spend time doing things that are not that high a priority. An alternative would be to create a link to the point in the article that is closest to a symbol.

pirapira:
@jamesray1 I doubt that the symbols in Wikipedia are the same as the symbols in YP.

jamesray1:
At any rate it's better to include the link until an actual table is developed.

pirapira:
@jamesray1 I don't believe that either.

Mitumaru (approved):

PR #669, opened 2018-03-09
Document what the SNARKV precompiled contract is doing

ValarDragon:
The yellow paper currently does not describe what snark precompiled contract is doing. I add documentation for what it is actually doing in this commit.


Paper.tex
```diff
@@ -1596,6 +1596,13 @@ \subsection{zkSNARK Related Precompiled Contracts}
 \end{eqnarray}
 We define $G_2$ to be the subgroup of $(C_2,+)$ generated by $P_2$. $G_2$ is known to be the only cyclic group of order $q$ on $C_2$. For a point $P$ in $G_2$, we define $\log_{P_2}(P)$ be the smallest natural number $n$ satisfying $n\cdot P_2=P$. With this definition, $\log_{P_2}(P)$ is at most $q-1$.
 
+Let $G_T$ be the multiplicative abelian group underlying $F_{p^{12}}$. It is known that a non-degenerate bilinear map $e : G_1\times G_2 \to G_T$ exists.
+This bilinear map is a type three pairing. The ZK-Snark verification contract takes a set of points $a_i$ that are in $G_1$, and a set of points $b_i$ that are in $G_2$. It checks that
+\begin{equation}
+\prod_{i=0}^{n}e\left(a_i, b_i\right) = 1
+\end{equation}
+If this is true, the contract returns $1$, otherwise it returns $0$, unless any of the points $a_i$ or $b_i$ are not on their respective curves, in which case the transaction is reverted.
+
```
pirapira: Please merge this description and the existing description around line 1644.

ValarDragon: Does the second commit I made address this satisfactorily?

pirapira:
Now \Xi_{\mathtt{SNARKV}} has two definitions.  YP usually doesn't do it.

ValarDragon:
Would you suggest any other way of mentioning the pairing check? It feels unclear from the pre-amble what the ZK-Snark contract is actually doing, whereas the precompiles for G1 addition and scalar multiplication are defined explicitly earlier. (Line 1559) Would it be better to instead say that we can verify that `log(a_1)*log(a_2)*... *log(a_n) * log(b_1)*log(b_2)*... *log(b_n) = 1` through using the bilinear map, and just say that the zk-snark verification contract is doing that operation? (I.e. the same style of statement that is used for the precompiled contract for addition on G_1)

Or do you think its not appropriate to mention that the contract is actually doing a pairing check?

pirapira:
@ValarDragon Of course, it's nicer to explain what is supposed to happen.  Then, it's important to make sure that two different descriptions match.  YP doesn't need to contain the proof that the two descriptions match (a reference to an external reference is ok).

YP does not need to say how the value is computed.  The most important thing is to specify what value should be computed without ambiguity.

ValarDragon:
Thank you for the response, that makes sense. I'll try to write up something later today that can better merge the two definitions. The reason I want to mention that the contract is using the pairing check is because this can be used for other pairing based cryptography applications such as BLS signatures. (Link to one such repo: https://github.com/Project-Arda/bgls-on-evm )

pirapira:
I'm glad that somebody well-versed in cryptography is looking at this.

ValarDragon:
I think my latest commit fixes there being multiple definitions, by indicating their equivalency. It also addresses how the pairing is being used in the contract, which addresses my original concern for providing documentation for people trying to build pairing based cryptography applications.

pirapira:
I'll have a look today.

Paper.tex
```diff
@@ -1596,6 +1596,14 @@ \subsection{zkSNARK Related Precompiled Contracts}
 \end{eqnarray}
 We define $G_2$ to be the subgroup of $(C_2,+)$ generated by $P_2$. $G_2$ is known to be the only cyclic group of order $q$ on $C_2$. For a point $P$ in $G_2$, we define $\log_{P_2}(P)$ be the smallest natural number $n$ satisfying $n\cdot P_2=P$. With this definition, $\log_{P_2}(P)$ is at most $q-1$.
 
+Let $G_T$ be the multiplicative abelian group underlying $F_{q^{12}}$. It is known that a non-degenerate bilinear map $e : G_1\times G_2 \to G_T$ exists.
```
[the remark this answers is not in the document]
ValarDragon: I added this in the latest commit to fix the parenthesis!

PR #670, opened 2018-03-09
Add braces to mathbf

ValarDragon:
This commit makes the bracing around the argument to mathbf as used in the precompiled contract section match the style used in the rest of the paper.


DanielMReed:
This shouldn't changed the output PDF, only the source, yes?

ValarDragon:
Yes

PR #678, opened 2018-04-01
Fix typo formula no.235

4000D:
This formula describes when two points are equivalent. So second point should be X_2, Y_2 for equality test


DanielRX:
Technically still correct as it was, given they are the same, (X_1,Y_1)+(X_2,Y_2) = (X_1,Y_1)+(X_1,Y_1).

PR #684, opened 2018-04-13
Correct transaction execution gas used equation

Matthalp:
This commit seeks to address a disconnect between the Yellow Paper
specification and what is implemented in the Ethereum clients in
practice for calculating the correct transaction receipt field. This
deviation has been verified in both [`cpp-ethereum`](https://github.com/ethereum/go-ethereum/blob/3caf16b15f0b6a30717acb245fa8d347b2f06c3f/core/state_transition.go#L236) and [`geth`](https://github.com/ethereum/cpp-ethereum/blob/aea3bc120761fe27fc696d0649ba5ae7a56a7915/libethereum/Executive.cpp#L499).

Currently the Yellow Paper specifies that the total gas by a
transaction is the difference between the transaction gas limit
and the gas remaining from the root-level message call or contract
creation within the transaction.

This neglects the fact that the transaction sender can be refunded
additional gas after the root-level message call or contract creation
has completed, which is problematic because this is used to compute
the cumulative gas used field for the transaction's corresponding
receipt in practice.

### DIFF
--- a/Paper.tex
+++ b/Paper.tex
@@ -643,12 +643,12 @@ and $T_{\mathrm{o}}$ is the original transactor, which can differ from the sende
 
 Note we use $\Theta_{4}$ and $\Lambda_{4}$ to denote the fact that only the first four components of the functions' values are taken; the final represents the message-call's output value (a byte array) and is unused in the context of transaction evaluation.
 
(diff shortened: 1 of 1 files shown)


Matthalp:
Thanks @pirapira !

PR #694, opened 2018-05-14
Remove "kern 10pc" visible in the PDF

pirapira:
A LaTeX formatting command was left visible in the PDF.  Moreover, the formatting command contained some concrete length values that can be easily removed from the source as well.


yperbasis:
Already fixed by another PR.

PR #695, opened 2018-05-22
Replace remaining \mathbb{P} with \mathbb{N}.

acoglio:
It looks like, when \mathbb{P} was replaced with \mathbb{N} in most of the
paper, three occurrences were not replaced. This commit replaces them.

Some trailing whitespace was also automatically removed.


pirapira:
Looks good to me.

PR #700, opened 2018-05-25
Clarify that a byte is 8 bits.

acoglio:
This is quite often the case, but not always, e.g. see ISO/IEC 9899:2011
standard for C. This added parenthetical remark avoids any possible confusion.


pirapira:
I agree.

PR #713, opened 2018-08-04
Fix Epirme

feiqihang:
I think the result of x/y should be a prime number.

### DIFF
--- a/Paper.tex
+++ b/Paper.tex
@@ -2435,7 +2435,7 @@ The size of the cache, $c_{size}$, is calculated using
 \end{equation}
 \begin{equation}
  E_{\mathrm{prime}}(x, y) = \begin{cases}
-x & \text{if} \quad x / y \in \mathbb{N} \\
+x & \text{if} \quad x / y \in \mathbb{P} \\
 E_{\mathrm{prime}}(x - 2 \cdot y, y) & \text{otherwise}
 \end{cases}
 \end{equation}


feiqihang:
To this end, I specifically looked at the source code of the go-ethereum project and the ethereumj project.
go-ethereum/consensus/ethash/algorithm.go:
`// calcCacheSize calculates the cache size for epoch. The cache size grows linearly,
// however, we always take the highest prime below the linearly growing threshold in order
// to reduce the risk of accidental regularities leading to cyclic behavior.
func calcCacheSize(epoch int) uint64 {
	size := cacheInitBytes + cacheGrowthBytes*uint64(epoch) - hashBytes
	for !new(big.Int).SetUint64(size / hashBytes).ProbablyPrime(1) { // Always accurate for n < 2^64
		size -= 2 * hashBytes
	}
	return size
}`

org.ethereum.mine.EthashParams:
`    /**
     * The parameters for Ethash's cache and dataset depend on the block number.
     * The cache size and dataset size both grow linearly; however, we always take the highest
     * prime below the linearly growing threshold in order to reduce the risk of accidental
     * regularities leading to cyclic behavior.
     */
    public long getCacheSize(long blockNumber) {
        long sz = CACHE_BYTES_INIT + CACHE_BYTES_GROWTH * (blockNumber / EPOCH_LENGTH);
        sz -= HASH_BYTES;
        while (!isPrime(sz / HASH_BYTES)) {
            sz -= 2 * HASH_BYTES;
        }
        return sz;
    }`

vasilevskykv:
I would like to ask one more question. Why in function PoW, construction sum from i=0 to nmix is used?
The term **sh(h,n)** doesn't depend on i. And we can calculate that nmix = 128/64 = 2. Thus, we can write not the sum, but **3 sh(h,n)**

Mitumaru (approved):

PR #719, opened 2018-11-07
Remove last reference to deprecated suicide op

KronicDeth:
# Changelog
## Bug Fixes
* [EIP-6](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-6.md) renamed `SUICIDE` to `SELFDESTRUCT`, but there was still a reference to it, which was confusing as nothing else in the paper explains suicide was the old name for selfdestruct.


nicksavers:
Thanks

PR #741, opened 2019-04-18
Fix font and improve wording.

acoglio:
The \mathrm{x} used as subscript to \mathbb{N} was not matching the plain
(\mathit) x in the phrase.

The wording was also not very clear. The new wording should be better.

This changes
old
to
new


acoglio:
@nicksavers Thank you for merging #742. Would you be willing to merge this one too?

PR #747, opened 2019-05-06
Treat t argument of HP (and f) as boolean.

acoglio:
The paragraph just before eq. (186) says that the second argument, t, of the
function HP, is a boolean value. The function HP is passed the boolean values
true and false as t in eq. (194). However, eq. (187) was treating t as a number,
in the condition 't \neq 0'. Even though booleans are isomorphic to {0, 1}, it's
clearer to change that condition to 't = true'.

This changes
old-hp-bool
to
new-hp-bool

Unless there are objections, I'll merge in a few days.

### DIFF
--- a/Paper.tex
+++ b/Paper.tex
@@ -1404,7 +1404,7 @@ It is defined as the function $\mathtt{HP}$ which maps from a sequence of nibble
 (16(f(t) + 1) + \mathbf{x}[0], 16\mathbf{x}[1] + \mathbf{x}[2], 16\mathbf{x}[3] + \mathbf{x}[4], ...) &
 \text{otherwise}
 \end{cases} \\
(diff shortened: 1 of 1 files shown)


nicksavers:
It might be good to leave this flag as a bit, as this is also the way it's programmed in [Geth](https://github.com/ethereum/go-ethereum/blob/master/trie/encoding.go#L29) and the [pseudocode](https://github.com/ethereum/wiki/wiki/Patricia-Tree#specification-compact-encoding-of-hex-sequence-with-optional-terminator).

acoglio:
I just made an alternative PR #749 that consistently treats t as a bit. Whether we use booleans or bits for this, the YP should be consistent.

acoglio:
I'm closing this PR because consensus was to go with #749.

PR #752, opened 2019-05-13
Paper.tex: correct layout of input data to BN_MUL

### LINKED ISSUE #751: incorrect input layout in BN_MUL definition
The layout of the input data to the `BN_MUL` zk-SNARK precompile is incorrect. Currently, it looks like this:

screenshot620

which means that the first 32 bytes of the input memory range are occupied by the scalar `n`, and the next 64 bytes are occupied by the curve point. In fact, according the Ethereum test suite and to other implementations like `aleth`, `parity`, and KEVM, the first 64 bytes are the curve point and the next 32 are the scalar, so it should look like this:

screenshot621

*correct implementation references:*
- [aleth](https://github.com/ethereum/aleth/blob/32f9c88f65fc30d7b682de651e2d40d8b7e26078/libdevcrypto/LibSnark.cpp#L192)
- [parity](https://github.com/paritytech/parity-ethereum/blob/2cf0d7d7757829a8d6d3a65547c96258c73b9b91/ethcore/src/builtin.rs#L467)
- [KEVM](https://github.com/kframework/evm-semantics/blob/master/evm.md#precompiled-contracts)
```
    syntax PrecompiledOp ::= "ECMUL"
 // --------------------------------
    rule <k> ECMUL => #ecmul((#asWord(DATA [ 0 .. 32 ]), #asWord(DATA [ 32 .. 32 ])), #asWord(DATA [ 64 .. 32 ])) ... </k>
         <callData> DATA </callData>
```

livnev:
fixes #751

### DIFF
--- a/Paper.tex
+++ b/Paper.tex
@@ -1697,8 +1697,8 @@ We define a precompiled contract for scalar multiplication on $G_1$, where $\bar
 \Xi_{\mathtt{BN\_MUL}}(\boldsymbol\sigma,g,I)&=&\left(\varnothing,0,A^0,()\right)\quad\text{if}\ x=\varnothing\\
 g_{\mathrm{r}} &=& 40000\\
 \mathbf{o}&\equiv&\delta_1^{-1}(n\cdot x)\quad\text{where $\cdot$ is the scalar multiplication in $G_1$}\\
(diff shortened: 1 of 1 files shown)


kriszrtina:
H

PR #765, opened 2019-08-11
Fix mathematical symbols.

feiqihang:
This should be the norm of o (|| o ||), not the absolute value (| o |).
This changes the following
image
into
image


acoglio:
Note: this also fixes the occurrence in eq. (97).
Screen Shot 2019-08-11 at 10 48 25 AM

PR #768, opened 2019-10-10
Fix the SHA3 instruction gas calculation.

feiqihang:
The description of <code>G<sub>sha3word</sub></code> is :
> Paid for each word (rounded up) for input data to a SHA3 operation.

According to the description of the `SHA3` instruction, we know that the second parameter in the stack <code>μ<sub>s</sub>[1]</code> is the number of words of the input data.
image

So, <code>G<sub>sha3word</sub></code> just a simple multiplication calculation.

This changes the following
image
to
image


yperbasis:
Already fixed by https://github.com/ethereum/yellowpaper/commit/2f60f3f548e5422a67112f8fc340f0c3ebb4fc5a

PR #772, opened 2019-11-16
Prefer MUST to SHOULD for requirements

fulldecent:


acoglio:
That seems appropriate, since the assertion (as I understand it) is an absolute requirement, not a suggestion or best practice.

The footnote contains another instance of 'should' (referred to CODESIZE), which I guess could be similarly changed to 'must'.

Mitumaru (approved):

fulldecent:
Requesting review please from @stehled

stehled:
@fulldecent I've reviewed the change, but I am not aware of the implementation details and therefore I don't feel like I can bring valuable insight on this matter.

Issue #767, opened 2019-09-04
MSIZE after access to offset ``2^256-1``


ekpyron:
If I see that correctly, then after ``MSTORE`` or ``MLOAD`` to ``2^256-1`` (no prior memory writes), ``μ_i'`` will be set to ``max(0,ceil(2^256 - 1 + 32)÷32))``, where the addition is specifically excluded from modulo arithmetic. That's ``2^251+1``.

Now the next ``MSIZE`` will return ``32μ_i`` with arithmetic modulo ``2^256`` - that's ``32``.

Wouldn't the more reasonable value be ``2^256-1``?

This is a rather academic corner case, since no client will implement this case, since it will always run out of gas anyways - but we are implementing an EVM interpreter for fuzzing the optimizer of the solidity compiler, which is used to check for execution trace equivalence *without* considering gas and thereby noticed this oddity (https://github.com/ethereum/solidity/pull/7342). What do you think about changing this in the yellow paper?

acoglio:
I haven't studied this part of the YP in detail yet, but it looks like perhaps the specifications of MLOAD and MSTORE should use min(µ<sub>s</sub>[0] + 32, 2<sup>256</sup>) instead of (µ<sub>s</sub>[0] + 32). That is, if µ<sub>s</sub>[0] + 32 exceeds 2<sup>256</sup>, and the 32 bytes starting at µ<sub>s</sub>[0] "wrap around" to memory locations 0 etc., we want to count all 2<sup>256</sup> bytes of memory as used: the min operation caps the number of used bytes at 2<sup>256</sup>. Then we should have the invariant µ<sub>i</sub> ≤ 2<sup>251</sup>.

There would still be a bit of a problem with MSIZE when all 2<sup>256</sup> memory bytes are used, because in this case µ<sub>i</sub> = 2<sup>251</sup> and 32µ<sub>i</sub> = 2<sup>256</sup>, which doesn't fit into a word. Perhaps returning 2<sup>256</sup>-1 is the best we can do in this case, but it would have to be a special case. In all other cases, MSIZE returns a multiple of 32.

(MSTORE8 looks unproblematic because, given the expected invariant µ<sub>s</sub>[0] < 2<sup>256</sup>, i.e. that µ<sub>s</sub>[0] is a word, we have µ<sub>s</sub>[0] + 1 ≤ 2<sup>256</sup>.)

ekpyron:
[...]

PR #775, opened 2020-06-03
Improve spec of MLOAD, MSTORE, and MSIZE.

acoglio:
This is in reference to Issue #767, which has all the relevant discussion.

This changes
old-mload-mstore
and
old-msize
to
new-mload-mstore
and
new-msize

### DIFF
--- a/Paper.tex
+++ b/Paper.tex
@@ -2146,12 +2146,12 @@ Here given are the various exceptions to the state transition rules given in sec
 \midrule
 0x51 & {\small MLOAD} & 1 & 1 & Load word from memory. \\
 &&&& $\boldsymbol{\mu}'_{\mathbf{s}}[0] \equiv \boldsymbol{\mu}_{\mathbf{m}}[\boldsymbol{\mu}_{\mathbf{s}}[0] \dots (\boldsymbol{\mu}_{\mathbf{s}}[0] + 31) ]$ \\
-&&&& $\boldsymbol{\mu}'_{\mathrm{i}} \equiv \max(\boldsymbol{\mu}_{\mathrm{i}}, \ceil{ (\boldsymbol{\mu}_{\mathbf{s}}[0] + 32) \div 32 })$ \\
+&&&& $\boldsymbol{\mu}'_{\mathrm{i}} \equiv \max(\boldsymbol{\mu}_{\mathrm{i}}, \ceil{ (\min(\boldsymbol{\mu}_{\mathbf{s}}[0] + 32, 2^{256}) \div 32 })$ \\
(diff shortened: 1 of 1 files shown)


Paper.tex
```diff
@@ -2186,7 +2186,7 @@ \subsection{Instruction Set}
 &&&& $\boldsymbol{\mu}'_{\mathbf{s}}[0] \equiv \boldsymbol{\mu}_{\mathrm{pc}}$ \\
 \midrule
 0x59 & {\small MSIZE} & 0 & 1 & Get the size of active memory in bytes. \\
-&&&& $\boldsymbol{\mu}'_{\mathbf{s}}[0] \equiv 32\boldsymbol{\mu}_{i}$ \\
+&&&& $\boldsymbol{\mu}'_{\mathbf{s}}[0] \equiv \begin{cases} 32\boldsymbol{\mu}_{i} & \text{if} \quad \boldsymbol{\mu}_{i} < 2^{251} \\ 2^{251}-1 & \text{otherwise} \end{cases}$ \\
```
ekpyron, line 2189:
```suggestion
&&&& $\boldsymbol{\mu}'_{\mathbf{s}}[0] \equiv \begin{cases} 32\boldsymbol{\mu}_{i} & \text{if} \quad \boldsymbol{\mu}_{i} < 2^{251} \\ 2^{256}-1 & \text{otherwise} \end{cases}$ \\
```
It's the size *in bytes* for "full memory usage", so ``2^256-1``, not ``2^251-1``, right?

acoglio:
@ekpyron Yes, definitely, thanks for catching it.

acoglio:
@nicksavers Does this look good to you too? If you confirm, you or I can merge it. Thanks.

nicksavers:
This limit has been a topic of discussion before. 

There is [EIP-1985: Sane limits for certain EVM parameters ](https://eips.ethereum.org/EIPS/eip-1985) and Geth (and some others) have already implemented a different value for MSIZE. You are free to take those values for your interpreter or add your input in the discussion thread. 

I suggest not to merge this until the EIP is Final.

acoglio:
@nicksavers Waiting for the EIP to be final is fine with me.

(We could still fix the overflow problem found by @ekpyron for the time being, even though it's a theoretical case, and then further revise things when the EIP is final. But perhaps we want to avoid too many revisions, especially if the EIP may be finalized soon(?).)

ekpyron:
IMHO we should merge and fix this independently of that EIP.

axic:
@nicksavers I think this is independent from EIP-1985. This PR fixes an ambiguity, while EIP-1985 introduces limits (which I think we should definitely be applied to the YP once those agreed).

Mitumaru (approved):

PR #776, opened 2020-06-09
Update authors.

paulmillr:
I think it makes sense to amend authors with new names at this point. See https://twitter.com/paulmillr/status/1270270652738605056


nicksavers:
Attribution will not be updated, unless the authors explicitly ask for it.

PR #777, opened 2020-06-18
[FIX] Status in receipt RLP may be != 0

yperbasis:
Please refer to [EIP-658](https://eips.ethereum.org/EIPS/eip-658).
Also check the go-ethereum [ implementation](https://github.com/ethereum/go-ethereum/blob/master/core/types/receipt.go#L36).

Change
Screenshot 2020-06-30 at 15 51 04
to
Screenshot 2020-06-30 at 15 51 27


nicksavers:
The `ReceiptStatusFailed` equals 0.

yperbasis:
> The `ReceiptStatusFailed` equals 0.

Yep, and `ReceiptStatusSuccessful` equals 1.

acoglio (approved):
I haven't studied this part of the YP in detail, but the proposed change should be fine.

PR #778, opened 2020-06-29
Corrections for CALLCODE and DELEGATECALL

yperbasis:
CALLCODE may be readonly executed when value != 0:
Since the sender and the recipient are the same, no value transfer actually happens: see "unless s = r" at the definition of sigma1 in Section 8 "Message Call".
go-ethereum does not throw `ErrWriteProtection` for CALLCODE with non-zero value, only for CALL:
https://github.com/ethereum/go-ethereum/blob/v1.10.3/core/vm/interpreter.go#L236

Change
Screenshot 2020-06-30 at 15 39 06
to
Screenshot 2020-06-30 at 15 37 52

There's no need to check balance in DELEGATECALL as there's no value transfer.
go-ethereum has this check for CALLCODE, but not for DELEGATECALL. Compare
https://github.com/ethereum/go-ethereum/blob/v1.10.3/core/vm/evm.go#L296
with
https://github.com/ethereum/go-ethereum/blob/v1.10.3/core/vm/evm.go#L327

Change
Screenshot 2020-06-30 at 15 44 14
to
Screenshot 2020-06-30 at 15 43 37

CALLCODE & DELEGATECALL don't bump nonce unlike CREATE, so they should use sigma rather than sigma* in line with CALL.

Change
Screenshot 2020-06-30 at 15 45 23
to
Screenshot 2020-06-30 at 15 45 38


chfast:
> CALLCODE may be readonly executed when value != 0

Did you make sure this case is covered by JSON tests?

yperbasis:
> > CALLCODE may be readonly executed when value != 0
> 
> Did you make sure this case is covered by JSON tests?

Yes, this case is well covered by the consensus tests, for example by [static_callcallcallcode_001_2](https://github.com/ethereum/tests/blob/develop/src/GeneralStateTestsFiller/stStaticCall/static_callcallcallcode_001_2Filler.json)

PR #779, opened 2020-07-03
Clarify reward rounding

yperbasis:
Currently it's not clear how to do the Wei rounding for block rewards.

Change
Screenshot 2020-07-03 at 17 35 26
to
Screenshot 2020-07-03 at 17 35 41

See go-ethereum implementation:
https://github.com/ethereum/go-ethereum/blob/v1.10.3/consensus/ethash/consensus.go#L641


yperbasis:
Doesn't really matter since R_block is divisible by 8 and by 32.

PR #781, opened 2020-07-25
Fix hyperlink

minaminao:
Hello. The transaction receipt hyperlink is "Transaction_Receipt", not "Transaction Receipt" :)

By the way, the hyperlink "Upsilon_pow_g2" also doesn't exist and warns. I don't know if this is normal, so I haven't fixed it in this pull request.


acoglio:
Looks good to me.

PR #787, opened 2021-03-15
Amend EXP gas cost inconsistency

fvictorio:
There is an inconsistency in how the gas cost of the EXP opcode is explained:

image

image

The first one uses a `ceil` while the second one uses a `floor`. I *think* the ceil one is the correct one (or at least that seems to be what `ethereumjs-vm` does). But in any case, they should be the same in both lines.


yperbasis (changes requested):
`1+floor(log_256(exponent))` is correct, it gives the number of bytes in the exponent.

For example, consider the case of exponent = 2:
`1+floor(log_256(exponent))` returns 1 (correct, the number 2 occupies 1 byte),
 while 
`1+ceil(log_256(exponent))` returns 2 (incorrect).

I suggest to change the comment from
Partial payment when multiplied by floor(log256(_exponent_)) for the EXP operation.
to
Partial payment when multiplied by the number of bytes in the _exponent_ for the EXP operation.

fvictorio:
Oh, I didn't notice the `1+`, that makes sense. I updated the PR with your suggestion.

PR #789, opened 2021-04-11
Update `SSTORE` cost and refund for EIP-2200

### LINKED ISSUE #788: SSTORE cost not updated for EIP-2200
[EIP-2200] changes the dynamic cost of `SSTORE` to refund some gas when
setting a storage location to its original value: i.e., the value to
which it would be set if the transaction reverted. But the yellow paper
still defines `C_{SSTORE}` only in terms of whether the storage location
is being set/reset to zero.

I poked around at adding this, but couldn't find prior art for notation
about the original state. Something like `\boldsymbol{\sigma}^0`?

It would at least be helpful to have a note in English referring the
reader to EIP-2200 for more details, as we do for `CREATE2` (EIP-1014).

EIP-2200 also changed the cost of `SLOAD`, but this is already handled
thanks to #783, since EIP-1884 made the same change.

[EIP-2200]: https://eips.ethereum.org/EIPS/eip-2200

wchargin:
This patch ports the decision tree from the specification of [EIP-2200]
into the instruction set reference. Fixes #788.

[EIP-2200]: https://eips.ethereum.org/EIPS/eip-2200

wchargin-branch: eip-2200

### DIFF
--- a/Paper.tex
+++ b/Paper.tex
@@ -2200,13 +2200,33 @@ Here given are the various exceptions to the state transition rules given in sec
 \midrule
 \linkdest{SSTORE}{}0x55 & {\small SSTORE} & 2 & 0 & Save word to storage. \\
 &&&& $\boldsymbol{\sigma}'[I_{\mathrm{a}}]_{\mathbf{s}}[ \boldsymbol{\mu}_{\mathbf{s}}[0] ] \equiv \boldsymbol{\mu}_{\mathbf{s}}[1] $ \\
-&&&&\linkdest{C__SSTORE}{} $C_{\text{\tiny SSTORE}}(\boldsymbol{\sigma}, \boldsymbol{\mu}) \equiv \begin{cases}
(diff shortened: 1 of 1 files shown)


wchargin:
Rendered:

![Screenshot of the new contents of the `SSTORE` row in the instruction set reference][ss]

[ss]: https://user-images.githubusercontent.com/4317806/114296508-99096600-9a60-11eb-83b2-acded684ea05.png

Implementation notes:

  - I’ve defined some local variables to improve readability: otherwise,
    it’s too hard to keep track of all the long chains of **σ**-indexes
    and verify that this matches the EIP.

  - I picked the name `\boldsymbol{\sigma}^0` by analogy with `A^0`, but
    would also be happy to use `\boldsymbol{\sigma}_0` or another name
    or formulation at maintainers’ discretion.

Generally open to suggestions; happy to incorporate feedback.

yperbasis:
@wchargin Thank a lot for this PR! I've made a clarification to make the cases mutually exclusive: https://github.com/ethereum/yellowpaper/pull/798

PR #801, opened 2021-06-14
EIP-3607: transactions from senders with deployed contract are forbidden

MariusVanDerWijden:
Draft PR for EIP-3607 (https://eips.ethereum.org/EIPS/eip-3607).
Waits for 3607 to be approved by the Core Devs

cc @dankrad

### DIFF
--- a/Paper.tex
+++ b/Paper.tex
@@ -281,7 +281,7 @@ An account is \textit{dead} when its account state is non-existent or empty:
 
 \subsection{The Transaction} \label{subsec:transaction}
 
-A transaction (formally, $T$) is a single cryptographically-signed instruction constructed by an actor externally to the scope of Ethereum. While it is assumed that the ultimate external actor will be human in nature, software tools will be used in its construction and dissemination\footnote{Notably, such `tools' could ultimately become so causally removed from their human-based initiation---or humans may become so causally-neutral---that there could be a point at which they rightly be considered autonomous agents. \eg contracts may offer bounties to humans for being sent transactions to initiate their execution.}. There are two types of transactions: those which result in message calls and those which result in the creation of new accounts with associated code (known informally as `contract creation'). Both types specify a number of common fields:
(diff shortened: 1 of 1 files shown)


holiman (approved):
LGTM (text-wise, haven't reviewed the rendered diff)

yperbasis (changes requested):
The condition should be
```
\boldsymbol{\sigma}[S(T)]_{\mathrm{c}} = \texttt{KEC}\big( () \big)
```
i.e. that sender's **codeHash** is that of an empty string.
For example, check the definition of an empty account in Eq (14).


`I_{\mathbf{b}}` isn't correct: it refers to initialising code, not sender's code.

Also, please add the condition to the validity Eq (59).

MariusVanDerWijden:
@yperbasis we discussed the change on ACD and decided to go with empty codehash as the condition instead of codesize as you proposed. Will un-draft the PR once the changes are merged into the EIP, the tests and the EIP is moved into Final

holiman:
See https://github.com/ethereum/go-ethereum/pull/23002#issuecomment-872282874 

Basically, a client implementing this change must be careful. If the same 'check' as `EXTCODEHASH` is reused for this check, then it needs to _allow_ both `0x000..` and `emptyCodeHash`. Because a transaction from an account that did not previously exist _is_ historically OK, if a miner accepts a gasPrice of zero. Which has happened.

yperbasis:
> See [ethereum/go-ethereum#23002 (comment)](https://github.com/ethereum/go-ethereum/pull/23002#issuecomment-872282874)
> 
> Basically, a client implementing this change must be careful. If the same 'check' as `EXTCODEHASH` is reused for this check, then it needs to _allow_ both `0x000..` and `emptyCodeHash`. Because a transaction from an account that did not previously exist _is_ historically OK, if a miner accepts a gasPrice of zero. Which has happened.

Should we include EIP-3607 into London or Shanghai then instead of applying it retroactively?

yperbasis (approved):

yperbasis:
Shall we merge it, given that it's already in geth (https://github.com/ethereum/go-ethereum/pull/23303) ?

PR #802, opened 2021-06-14
Amend contract creation when initialising code reverts

yperbasis:
Change

to


acoglio (approved):
Should be fine, but i haven't studied this part of the YP in detail.

PR #806, opened 2021-06-21
Specify the behaviors of EXTCODESIZE and EXTCODECOPY when the target …

yperbasis:
…account does not exist

Change

to


holiman:
The behavious of `EXTCODESIZE` is IMO pretty well-defined. The size of something that does not exist is zero. But sure, I don't object

yperbasis:
> The behavious of `EXTCODESIZE` is IMO pretty well-defined. The size of something that does not exist is zero. But sure, I don't object

Typically the Yellow Paper goes an extra mile to handle non-existing accounts. See, for example, Eq (86). To my mind accessing the nonce (or the codeHash) of an non-existing account is a mathematical equivalent of dereferencing a null pointer.

PR #808, opened 2021-07-05
Remove duplicated T_t

stanbar:
The T_t seems to be duplicated in this tuple.


bendyarm (changes requested):
I think the T_t is supposed to be duplicated.  See p.10:
image

stanbar:
You are right.

PR #820, opened 2021-10-15
Fix bit-selection in bloom filter function

### LINKED ISSUE #819: bloom filter has incorrect semantics
### Notation
The bloom filter function (BFF) `M` is defined as the bit-wise OR of a set of elements run through the `M_{3:2048}` function:

drawing

I'll call the former the `combined BFF` and the latter a `single BFF`.

### Background
The single BFF is defined as follows:

drawing

In plain English, it amounts to returning a bit sequence of length 2048 with up to 3 bits set to 1.

### Description
I believe the `m` function is defined differently than in most implementations.

**Let's have a look at K-EVM**

On one of last lines of https://github.com/kframework/evm-semantics/blob/master/evm.md#block-processing, we have:

```k
rule setBloomFilterBits(HASH) => (1 <<Int getBloomFilterBit(HASH, 0)) |Int (1 <<Int getBloomFilterBit(HASH, 2)) |Int (1 <<Int getBloomFilterBit(HASH, 4))

rule getBloomFilterBit(X, I) => #asInteger(X [ I .. 2 ]) %Int 2048
```

This implementation will set bits counting from the least-significant bit. E.g. if the first byte of the hash is `0x00`, then only the least-significant bit will be set. (Note that this implementation *may* be subject to a [different issue](https://github.com/kframework/evm-semantics/issues/1144).)

**Let's have a look at eth-bloom, which is used in [`py-evm`](https://github.com/ethereum/py-evm/blob/21759ee681315f7099b14893b6ac6d1a5e659bc0/setup.py#L9)**

The function [`chunk_to_bloom_bits`](https://github.com/ethereum/eth-bloom/blob/03378183fa3afefff0f366f1a8e9a31d56363b28/eth_bloom/bloom.py#L20-L22) is defined as such:

```python3
def chunk_to_bloom_bits(chunk: bytes) -> int:
    high, low = bytearray(chunk)
    return 1 << ((low + (high << 8)) & 2047)
```

This implementation is the same as K.

**Yellow Paper**

In the yellow paper however, `\mathcal{B}` counts from most-significant bits. Not only is this implied by a lack of specification thereof, it is also evident from further uses of this function:

- https://github.com/ethereum/yellowpaper/blob/6ef60627498bb63bf74b7ffb7a9d2ad207745afa/Paper.tex#L742
- https://github.com/ethereum/yellowpaper/blob/6ef60627498bb63bf74b7ffb7a9d2ad207745afa/Paper.tex#L1834
- https://github.com/ethereum/yellowpaper/blob/6ef60627498bb63bf74b7ffb7a9d2ad207745afa/Paper.tex#L1866

As a result, if the first byte of the hash is `0x00`, the most-significant bit will be set. This is at odds with at least two implementations, which activate the least-significant bit in such a case.

### Recommendation
Change the semantics of the `m(x, i)` function to be a complement of 2047. This will have the same effect as counting from least-significant bits, and so ensure correct semantics.

hacker-DOM:
Fixes #819

### DIFF
--- a/Paper.tex
+++ b/Paper.tex
@@ -426,7 +426,7 @@ where $M_{3:2048}$ is a specialised Bloom filter that sets three bits out of 204
 M_{3:2048}(\mathbf{x}: \mathbf{x} \in \mathbb{B}) & \equiv & \mathbf{y}: \mathbf{y} \in \mathbb{B}_{256} \quad \text{where:}\\
 \mathbf{y} & = & (0, 0, ..., 0) \quad \text{except:}\\
 \forall i \in \{0, 2, 4\}&:&\mathcal{B}_{m(\mathbf{x}, i)}(\mathbf{y}) = 1\\
-m(\mathbf{x}, i) &\equiv& \mathtt{KEC}(\mathbf{x})[i, i + 1] \bmod 2048
+m(\mathbf{x}, i) &\equiv& 2047 - (\mathtt{KEC}(\mathbf{x})[i, i + 1] \bmod 2048)
 \end{eqnarray}
 
 where $\mathcal{B}$ is the bit reference function such that $\mathcal{B}_{\mathrm{j}}(\mathbf{x})$ equals the bit of index $j$ (indexed from 0) in the byte array $\mathbf{x}$.
--- a/Paper.tex
+++ b/Paper.tex
@@ -429,7 +429,7 @@ M_{3:2048}(\mathbf{x}: \mathbf{x} \in \mathbb{B}) & \equiv & \mathbf{y}: \mathbf
 m(\mathbf{x}, i) &\equiv& 2047 - (\mathtt{KEC}(\mathbf{x})[i, i + 1] \bmod 2048)
 \end{eqnarray}
 
-where $\mathcal{B}$ is the bit reference function such that $\mathcal{B}_{\mathrm{j}}(\mathbf{x})$ equals the bit of index $j$ (indexed from 0) in the byte array $\mathbf{x}$.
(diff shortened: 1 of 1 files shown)


Paper.tex
```diff
@@ -426,7 +426,7 @@ \subsubsection{Transaction Receipt}\linkdest{Transaction_Receipt}{}
 M_{3:2048}(\mathbf{x}: \mathbf{x} \in \mathbb{B}) & \equiv & \mathbf{y}: \mathbf{y} \in \mathbb{B}_{256} \quad \text{where:}\\
 \mathbf{y} & = & (0, 0, ..., 0) \quad \text{except:}\\
 \forall i \in \{0, 2, 4\}&:&\mathcal{B}_{m(\mathbf{x}, i)}(\mathbf{y}) = 1\\
-m(\mathbf{x}, i) &\equiv& \mathtt{KEC}(\mathbf{x})[i, i + 1] \bmod 2048
+m(\mathbf{x}, i) &\equiv& 2047 - (\mathtt{KEC}(\mathbf{x})[i, i + 1] \bmod 2048)
 \end{eqnarray}
 
 where $\mathcal{B}$ is the bit reference function such that $\mathcal{B}_{\mathrm{j}}(\mathbf{x})$ equals the bit of index $j$ (indexed from 0) in the byte array $\mathbf{x}$.
```
yperbasis, line 432: I'd add a clarification here:
Within a byte $\mathcal{B}$ treats bits as big-endian (more significant bits have smaller indices).

yperbasis (approved):

Paper.tex
hacker-DOM: I agree! I've added that clarification, but for the entire byte-array

yperbasis (approved):

PR #822, opened 2021-11-01
Amendment: transaction sender might not exists

yperbasis:
Currently the Yellow Paper states that transactions from a non-existing sender are invalid. However, historically there've been such transactions with 0 gas price, for example https://etherscan.io/tx/0x0874e15fd07f92bf1143f9abd289c7c0737330782457b9c15d51666021b1194a.

Also, geth does allow such transactions (see [preCheck](https://github.com/ethereum/go-ethereum/blob/v1.10.11/core/state_transition.go#L217) and [statedb.go](https://github.com/ethereum/go-ethereum/blob/v1.10.11/core/state/statedb.go#L260)).

This change amends the Yellow Paper to account for such transactions.
Before:
Screenshot 2021-11-01 at 13 23 21 png
After:
Screenshot 2021-11-01 at 13 22 13 png

### DIFF
--- a/Paper.tex
+++ b/Paper.tex
@@ -655,7 +655,6 @@ The validity is determined as:
 \begin{equation}
 \begin{array}[t]{rcl}
 S(T) & \neq & \varnothing \quad \wedge \\
-\boldsymbol{\sigma}[S(T)] & \neq & \varnothing \quad \wedge \\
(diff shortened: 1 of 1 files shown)


holiman (approved):
LGTM

PR #825, opened 2021-11-25
Replace SHA3 opcode with KECCAK256

axic:
This was [proposed in 2016](https://github.com/ethereum/EIPs/issues/59), [Solidity uses this since 2017](https://github.com/ethereum/solidity/blob/develop/Changelog.md#0412-2017-07-03), and evmone and some other VMs use the `keccak256` name.

However since this was never updated in the Yellow Paper, notably geth (+ erigon) and nethermind still use the `sha3` name.

A similar renaming took place from `suicide` to `selfdestruct` which is the term Yellow Paper uses (and no reference is kept to `suicide`). I think the `keccak256` is more expressive, and given the previous case this renaming likely won't cause much harm.


acoglio:
Looks good.

yperbasis (approved):

PR #826, opened 2021-11-26
Use keccak-256 consistently

axic:
The document uses various forms to refer to Keccak-256:
- `Keccak-256`
- `Keccak 256-bit hash`
- `Keccak`
- plus the `KEC` function

However there is also the odd need for Keccak-512 / KEC512.

In this PR I have clarified all the cases where `Keccak` was used alone to state `Keccak-256`, but haven't modified the cases using `Keccak 256-bit hash`.


acoglio:
Looks good.

yperbasis (approved):

PR #827, opened 2021-11-26
Implement EIP-2681: Limit account nonce to 2^64-1

### LINKED ISSUE #813: Implement EIP-2681: Limit account nonce to 2^64-1
> Introduce two new restrictions retroactively from genesis:
>
> 1. Consider any transaction invalid, where the nonce exceeds `2^64-1`.
> 2. The `CREATE` and `CREATE2` instructions to abort with an exceptional halt, where the account nonce is `2^64-1`.

This was accepted on [ACD120](https://github.com/ethereum/pm/blob/master/All%20Core%20Devs%20Meetings/Meeting%20120.md#limiting-account-nonce-eip-2681-vs-3338).

axic:
Fixes #813.


Paper.tex
```diff
@@ -306,7 +306,7 @@ \subsection{The Transaction} \label{subsec:transaction}
...
 
 \begin{description}
-\item[nonce]\linkdest{tx_nonce}{} A scalar value equal to the number of transactions sent by the sender; formally $T_{\mathrm{n}}$.
+\item[nonce]\linkdest{tx_nonce}{} A 64-bit scalar value equal to the number of transactions sent by the sender; formally $T_{\mathrm{n}}$.
```
axic, line 309: Perhaps the "64-bit" mention here is not entirely correct, as it is serialised via RLP, so it has no fixed size. However since the recipient field below states "160-bit address" I decided to include it.

Potentially the more clear way is to include these restrictions in the validation sections, which means the address should be just a scalar here, and a new rule would need to be added.

acoglio:
Equations (13) and (17) say that the nonce is a 256-bit word. Should they be changed too?

axic:
Good point @acoglio.

I think (13) is worth changing as that already mentions "types" (such as `B_32` vs `N_256`) and doesn't deal with RLP.

However not sure about (17). That seems to deal with RLP encoded data and is less strict about types. It is also missing `T_t` which should be `N_160`.

Mitumaru (approved):

PR #828, opened 2021-11-30
corrected the definition of N_n on page 5 eq.(18)

suahk:
- Current version defines elements belonging to N_n has to be less than 2^n (i.e., in bits) => but it should be in bytes < (2^3)^n
- For example N_256 should represents elements that can be written in 256 bytes (not 256 bits)


yperbasis:
Nope, N_256 correctly represents 256-bit unsigned integers. A 256-bit unsigned integer fits perfectly into 32-byte EVM words.

PR #830, opened 2021-12-01
Genesis block difficulty is 2^34, not 2^17

yperbasis:
See https://etherscan.io/block/0

Before:
Screenshot 2021-12-01 at 12 06 28

After:
Screenshot 2021-12-01 at 12 06 44


acoglio (approved):
I haven't really studied this part of the Yellow Paper, but the changes looks fine as far as I can tell.

PR #836, opened 2021-12-27
Update Rsclear from 15000 to 4800

jrocco2:
Adjusting the value of Rsclear as suggested by [EIP-3529](https://eips.ethereum.org/EIPS/eip-3529)

Old:
image

New:
Screen Shot 2021-12-27 at 3 55 45 pm

### DIFF
--- a/Paper.tex
+++ b/Paper.tex
@@ -1913,7 +1913,7 @@ $G_{\mathrm{coldsload}}$ & 2100 & Cost of a cold storage access. \\
 $G_{\mathrm{sset}}$ & 20000 & Paid for an {\small SSTORE} operation when the storage value is set to non-zero from zero. \\
 $G_{\mathrm{sreset}}$ & 2900 & Paid for an {\small SSTORE} operation when the storage value's zeroness remains unchanged or\\
 &&is set to zero. \\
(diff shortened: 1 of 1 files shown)


yperbasis:
Currently the YP describes the Berlin version of the protocol (actually, [EIP-2718](https://eips.ethereum.org/EIPS/eip-2718) & [EIP-2930](https://eips.ethereum.org/EIPS/eip-2930) are still pending). We first need to finish updating the YP to Berlin before moving to London EIPs. Also, I would prefer to implement all of [EIP-3529](https://eips.ethereum.org/EIPS/eip-3529) in one PR, not just a part.

jrocco2:
Ah makes sense! Didn't realise Berlin was still in process. Is there a list of things that still need to be done perhaps I can contribute?

yperbasis:
> Ah makes sense! Didn't realise Berlin was still in process. Is there a list of things that still need to be done perhaps I can contribute?

Sure! I suggest to extend this PR to cover the entire EIP-3529.

Mitumaru (approved):

yperbasis:
Done by PR #892.

Issue #844, opened 2022-02-15
[Question] How to calculate the available gas after interpreting CALL opcode


ZhangZhuoSJTU:
I am a bit confused about the following formula in YP.

https://github.com/ethereum/yellowpaper/blob/630ed566d4d147124783bbad9d9d3aa5c3522e2b/Paper.tex#L2588

Specifically, (In appendix __H.2. Instruction Set.__) after evaluating a CALL opcode, the available gas is defined by the blue circled formula.

![image](https://user-images.githubusercontent.com/14835483/153975674-7adbe1a3-ac90-45f5-91d7-cd36fe56c059.png)

Based on my understanding, `\mu_g` denotes the available gas before evaluating the CALL opcode, `g^'` denotes the available gas of the callee function, and `\mu_g^'` denotes the available gas after evaluating the CALL opcode. Please kindly correct me if I made it wrong.

I am then a bit confused why it is an addition. I feel the correct version may be the black circled formula as follows. Here I only consider the case of the call being successfully processed, i.e., not the case of insufficient balance or exceeded call depth.

![image](https://user-images.githubusercontent.com/14835483/153976558-01feabc9-fede-4f03-99e7-4e46bf23ef66.png)

Besides, I failed to find the definition of the yellow circled `g`. 

![image](https://user-images.githubusercontent.com/14835483/153977122-1223977c-e317-4e39-8754-0881eb9baf7d.png)

I am not sure whether I understand it correctly. It would be very appreciated if anyone can shed some light on this.

Many thanks!

yperbasis:
You're right, the YP was misspecified. I've submitted PR #846.

PR #846, opened 2022-02-20
Fix gas specification in CALL* & CREATE* operations

yperbasis:
Issue #844 explains the problem.

Before:
before
After:
after

Before:
before
After:
after


acoglio (approved):
I haven't studied this part of the YP in great detail, but the change looks fine to me.

PR #847, opened 2022-02-21
Fix the upsidedown multiplier in the bloom filter footnote

rlgns98kr:
스크린샷 2022-02-21 오후 3 46 57
The operation of the bloom filter is explained in the image above.

I fix the upsidedown multiplier in the bloom filter footnote.
Before:
스크린샷 2022-02-21 오후 3 48 53
After:
스크린샷 2022-02-21 오후 3 46 10


rlgns98kr:
#848

PR #858, opened 2022-04-29
Specify how REVERT returns data the same as RETURN does

LEAFERx:


yperbasis:
Could you please attach an image rendering the added formula?

LEAFERx:
image
image

yperbasis (approved):

PR #859, opened 2022-05-01
Some minor updates

qbzzt:
1. Added the missing two `CALL` opcodes in section 9.2
2. Updated the memory expansion cost
3. Fixed the initial value in (142)


Paper.tex
```diff
@@ -1076,7 +1076,7 @@ \subsection{Execution Overview}
 \boldsymbol{\mu}_{\mathbf{m}} & \equiv & (0, 0, ...) \\
 \boldsymbol{\mu}_{\mathrm{i}} & \equiv & 0 \\
 \boldsymbol{\mu}_{\mathbf{s}} & \equiv & () \\
-\boldsymbol{\mu}_{\mathbf{o}} & \equiv & ()
+\boldsymbol{\mu}_{\mathbf{o}} & \equiv & $\varnothing$ \\
```
yperbasis, line 1079: **μ<sub>o</sub>** is different from **o**. The second paragraph of section 9.4 refers to **o**, not **μ<sub>o</sub>**. The latter is the output substate of machine state **μ** and thus should be initialized with an empty sequence, not ∅.

yperbasis (changes requested):

PR #862, opened 2022-06-09
Fix: revert message should be available in the returndata buffer in case a CREATE reverts

zzh1996:
To be consistent with [EIP-140](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-140.md?plain=1#L37) and [Geth implementation](https://github.com/ethereum/go-ethereum/blob/v1.10.18/core/vm/instructions.go#L619)


zzh1996:
Before:

image

After:

image

yperbasis:
Unfortunately, right now _z_ doesn't differentiate between `REVERT` and non-`REVERT` failures. I'm going to make a prerequisite change to the Yellow Paper so that `z` has 3 values:
0 - failure other than `REVERT`
1 - `REVERT`
2 - success

zzh1996:
> Unfortunately, right now _z_ doesn't differentiate between `REVERT` and non-`REVERT` failures. I'm going to make a prerequisite change to the Yellow Paper so that _z_ has 3 values: 0 - failure other than `REVERT` 1 - `REVERT` 2 - success

Why do we need to differentiate between REVERT and non-REVERT failures? For non-REVERT failures `o` will be empty and everything is correct, just like in the CALL opcode.

yperbasis:
> Why do we need to differentiate between REVERT and non-REVERT failures? For non-REVERT failures `o` will be empty and everything is correct, just like in the CALL opcode.

Yes, I think you're right, at least judging by Eq (154). Btw, I think that Section 9.4 "Execution Overview" is especially badly written and confusing 🤯

yperbasis (approved):

PR #870, opened 2022-07-13
Fix memory linear cost limit: 724B -> 704B

LCamel:
image It should be "linear up to 735B", because

image

The "a" in the C_mem function is number of *words*.
We have to do the ceiling/floor first, and convert it to bytes later.
```
>>> math.ceil(math.sqrt(512) * 32) - 1  # wrong formula
724.0

>>> math.ceil(math.sqrt(512)) * 32 - 1  # right formula
735.0
```
Verify:

The non-linear term becomes non-zero at 735/736B, not 724/725B.
```
>>> x = 724; (x / 32) ** 2 / 512
0
>>> x = 725; (x / 32) ** 2 / 512        # still 0, incorrect
0

>>> x = 735; (x / 32) ** 2 / 512
0
>>> x = 736; (x / 32) ** 2 / 512        # becomes non-zero, correct
1
```

### DIFF
--- a/Paper.tex
+++ b/Paper.tex
@@ -2111,7 +2111,7 @@ Note the memory cost component, given as the product of $G_{\mathrm{memory}}$ an
 
 Referencing a zero length range (e.g. by attempting to pass it as the input range to a CALL) does not require memory to be extended to the beginning of the range. $\boldsymbol{\mu}'_{\mathrm{i}}$ is defined as this new maximum number of words of active memory; special-cases are given where these two are not equal.
 
-Note also that $C_{\mathrm{mem}}$ is the memory cost function (the expansion function being the difference between the cost before and after). It is a polynomial, with the higher-order coefficient divided and floored, and thus linear up to 724B of memory used, after which it costs substantially more.
+Note also that $C_{\mathrm{mem}}$ is the memory cost function (the expansion function being the difference between the cost before and after). It is a polynomial, with the higher-order coefficient divided and floored, and thus linear up to 704B of memory used, after which it costs substantially more.
 
 While defining the instruction set, we defined the memory-expansion for range function, $M$, thus:
 


yperbasis (changes requested):
Let _x_ be the number of bytes and _a_ be the number of words,
```
a = math.ceil(x/32)
```
So the quadratic cost term is
```
math.floor( (math.ceil(x/32))**2 / 512)
```
and is 0 up to **704** bytes.

LCamel:
x = 0 -> a = 0
x = 1~32 -> a = 1
x = 33~64 -> a = 2
a = math.ceil((x + 31) / 32)

https://github.com/ethereum/go-ethereum/blob/d8ff53dfb8a516f47db37dbc7fd7ad18a1e8a125/core/vm/common.go#L72

yperbasis:
> https://github.com/ethereum/go-ethereum/blob/d8ff53dfb8a516f47db37dbc7fd7ad18a1e8a125/core/vm/common.go#L72

The formula corresponding to the geth implementation in not `a = math.ceil((x + 31) / 32)`, but `a = math.floor((x + 31) / 32)`, which is for integers the same as `a = math.ceil(x / 32)`

LCamel:
You are right. I typed the wrong formula in the reply.
By using a = math.ceil(x / 32), we will have
```
>>> x = 704; math.floor( (math.ceil(x/32))**2 / 512)
0.0
>>> x = 705; math.floor( (math.ceil(x/32))**2 / 512)
0.0
>>> x = 724; math.floor( (math.ceil(x/32))**2 / 512)
0.0
>>> x = 725; math.floor( (math.ceil(x/32))**2 / 512)
0.0
>>> x = 735; math.floor( (math.ceil(x/32))**2 / 512)
0.0
>>> x = 736; math.floor( (math.ceil(x/32))**2 / 512)
1.0
```
The quadratic cost term is 0 up to 735 bytes.

yperbasis:
> You are right. I typed the wrong formula in the reply. By using a = math.ceil(x / 32), we will have
> 
> ```
> >>> x = 704; math.floor( (math.ceil(x/32))**2 / 512)
> 0.0
> >>> x = 705; math.floor( (math.ceil(x/32))**2 / 512)
> 0.0
> >>> x = 724; math.floor( (math.ceil(x/32))**2 / 512)
> 0.0
> >>> x = 725; math.floor( (math.ceil(x/32))**2 / 512)
> 0.0
> >>> x = 735; math.floor( (math.ceil(x/32))**2 / 512)
> 0.0
> >>> x = 736; math.floor( (math.ceil(x/32))**2 / 512)
> 1.0
> ```
> 
> The quadratic cost term is 0 up to 735 bytes.

This is not true. Take x = 705.
⌈705/32⌉ = 23
23^2 = 529
⌊529/512⌋ = 1

LCamel:
The number has been corrected. Thank you for the help!

LCamel:
The "a" in the C_mem function is number of *words*.

If a * a >= 512, the cost will become non-linear.

1 word = 32 bytes
```
>>> bytes_used = 704
>>> a = int(math.ceil(float(bytes_used) / 32))
>>> a
22
>>> a * a >= 512
False

>>> bytes_used = 705
>>> a = int(math.ceil(float(bytes_used) / 32))
>>> a
23
>>> a * a >= 512
True
```
The cost is linear up to 704B.

yperbasis (approved):

PR #876, opened 2022-09-28
Fix DIFFICULTY opcode with PREVRANDAO

ethosdev:
Reference: https://eips.ethereum.org/EIPS/eip-4399

Since the Merge, opcode 0x44 returns the output of the randomness beacon provided by the beacon chain


yperbasis:
At the moment the Yellow Paper corresponds to the [Berlin](https://github.com/ethereum/execution-specs/blob/master/network-upgrades/mainnet-upgrades/berlin.md) upgrade. We first need to update it to [London](https://github.com/ethereum/execution-specs/blob/master/network-upgrades/mainnet-upgrades/london.md) before [The Merge](https://github.com/ethereum/execution-specs/blob/master/network-upgrades/mainnet-upgrades/paris.md).

PR #881, opened 2022-12-11
Tiny Correction for returndata in Create

ritzdorf:
Based on the definition of z above, where it is `z = 0` in the error case, I think it should be `z = 0` instead of `z = 1` in this part as well. Alternatively, the whole line could simplified as 
```
$\boldsymbol{\mu}'_{\mathbf{o}} = \mathbf{o}$
```


yperbasis:
This bit of the YP is not very clear, but I think the logic is that the output is returned in the case of `REVERT` and is _not_ returned in the case of success. See [EIP-140](https://eips.ethereum.org/EIPS/eip-140#specification) and https://github.com/ethereum/go-ethereum/blob/v1.10.26/core/vm/instructions.go#L618.

PR #902, opened 2024-02-20
Sentence made clearer and simpler

honzaprevratil:
I am not sure, that the "should" is correctly used in the sentence. Although it may be correct, it creates unnecessary confusion thanks to it's other meaning of (uncertain) expectancy, in my opinion.

If it gets replaced by word "when" the sentence has clearer meaning to me removing any doubt.


yperbasis:
"should" can be used in conditionals. See https://dictionary.cambridge.org/grammar/british-grammar/conditionals-other-expressions-unless-should-as-long-as

honzaprevratil:
OK, so it IS used correctly, but in my opinion, it still looks confusing in the middle of the sentence. Anyway, thanks for the reference.

PR #903, opened 2024-02-23
Transaction recipient should be warm

pldespaigne:
# TL;DR
|Before|After|
|-|-|
|image|image|

# Problem
The [Berlin](https://ethereum.org/history#berlin) hardfork introduced [EIP-2929](https://eips.ethereum.org/EIPS/eip-2929) which is now part of the Yellow Paper specification. This EIP states that the transaction recipient address should be part of the initial accessed account set:
image

The Yellow Paper **does** enforce this behavior **only** for contract creation and subsequent xCALL opcodes **but not for the initial message call**.

# Proposed Solution
When the transaction is a message call ($T_t \neq \varnothing$) add the recipient address ($T_t$) to the accessed account set, along with the already present precompiled, sender address, and access list.

---
I would be happy to discuss this further or to make some change to this PR if needed .
Anyway thanks for your time!

### DIFF
--- a/Paper.tex
+++ b/Paper.tex
@@ -825,8 +825,12 @@ Evaluating $\boldsymbol{\sigma}_{\mathrm{P}}$ from $\boldsymbol{\sigma}_0$ depen
 where
 \begin{align}
 A^* & \equiv A^0 \quad \text{except} \\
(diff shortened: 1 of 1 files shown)


pldespaigne:
Hey @yperbasis 👋
Thanks for the review!

I would be interested in updating the YP to the Shangai fork and next, to the Cancun fork, is anyone already working on this?
Do you think I could ask for a small EF grant?

pldespaigne:
Thanks!
I already applied, I'm waiting for them to answer now 🙂

PR #904, opened 2024-02-26
Typo in word memorise

honzaprevratil:
Fixes typo `memoise`  to `memorise`.


yperbasis:
"memoise" is a valid term. See https://en.wikipedia.org/wiki/Memoization

PR #909, opened 2024-06-10
Fix returndata details

RenanSouza2:
This includes the returndata cache in the description of the machine state 

and also corrects the definition of the behaviour of the returndatacopy opcode

### DIFF
--- a/Paper.tex
+++ b/Paper.tex
@@ -1187,7 +1187,7 @@ Note that, when we evaluate $\Xi$, we drop the fourth element $I'$ and extract t
 $X$ is thus cycled (recursively here, but implementations are generally expected to use a simple iterative loop) until either \hyperlink{zhalt}{$Z$} becomes true indicating that the present state is exceptional and that the machine must be halted and any changes discarded or until \hyperlink{hhalt}{$H$} becomes a series (rather than the empty set) indicating that the machine has reached a controlled halt.
 
 \subsubsection{Machine State}
-The machine state $\boldsymbol{\mu}$ is defined as the tuple $(g, pc, \mathbf{m}, i, \mathbf{s}, \mathbf{o})$ which are the gas available, the program counter $pc \in \mathbb{N}_{256}$ , the memory contents, the active number of words in memory (counting continuously from position 0), the stack contents, and the returndata cache. The memory contents $\boldsymbol{\mu}_{\mathbf{m}}$ are a series of zeroes of size $2^{256}$.
+The machine state $\boldsymbol{\mu}$ is defined as the tuple $(g, pc, \mathbf{m}, i, \mathbf{s}, \mathbf{o})$ which are the gas available, the program counter $pc \in \mathbb{N}_{256}$ , the memory contents, the active number of words in memory (counting continuously from position 0), the stack contents, and the returndata buffer. The memory contents $\boldsymbol{\mu}_{\mathbf{m}}$ are a series of zeroes of size $2^{256}$.
 
 For the ease of reading, the instruction mnemonics, written in small-caps (\eg \space {\small ADD}), should be interpreted as their numeric equivalents; the full table of instructions and their specifics is given in Appendix \ref{app:vm}.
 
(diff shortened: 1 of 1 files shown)


Paper.tex
```diff
@@ -1187,7 +1187,7 @@ \subsection{Execution Overview}
...
 
 \subsubsection{Machine State}
-The machine state $\boldsymbol{\mu}$ is defined as the tuple $(g, pc, \mathbf{m}, i, \mathbf{s})$ which are the gas available, the program counter $pc \in \mathbb{N}_{256}$ , the memory contents, the active number of words in memory (counting continuously from position 0), and the stack contents. The memory contents $\boldsymbol{\mu}_{\mathbf{m}}$ are a series of zeroes of size $2^{256}$.
+The machine state $\boldsymbol{\mu}$ is defined as the tuple $(g, pc, \mathbf{m}, i, \mathbf{o}, \mathbf{s})$ which are the gas available, the program counter $pc \in \mathbb{N}_{256}$ , the memory contents, the active number of words in memory (counting continuously from position 0), the returndata cache,and the stack contents. The memory contents $\boldsymbol{\mu}_{\mathbf{m}}$ are a series of zeroes of size $2^{256}$.
```
wjmelements, line 1190: It's a [buffer](https://eips.ethereum.org/EIPS/eip-211) not a [cache](https://en.wikipedia.org/wiki/Cache_(computing))

pldespaigne:
For context, here is a screenshot diff of the proposed changes

|before|after|
|-|-|
|0_before|0_after|
|1_before|1_after|

Paper.tex
```diff
@@ -1187,7 +1187,7 @@ \subsection{Execution Overview}
...
 
 \subsubsection{Machine State}
-The machine state $\boldsymbol{\mu}$ is defined as the tuple $(g, pc, \mathbf{m}, i, \mathbf{s})$ which are the gas available, the program counter $pc \in \mathbb{N}_{256}$ , the memory contents, the active number of words in memory (counting continuously from position 0), and the stack contents. The memory contents $\boldsymbol{\mu}_{\mathbf{m}}$ are a series of zeroes of size $2^{256}$.
+The machine state $\boldsymbol{\mu}$ is defined as the tuple $(g, pc, \mathbf{m}, i, \mathbf{o}, \mathbf{s})$ which are the gas available, the program counter $pc \in \mathbb{N}_{256}$ , the memory contents, the active number of words in memory (counting continuously from position 0), the returndata cache, and the stack contents. The memory contents $\boldsymbol{\mu}_{\mathbf{m}}$ are a series of zeroes of size $2^{256}$.
```
pldespaigne, line 1190: As previously said by @wjmelements it's a "buffer" and not a "cache", also I would suggest to put $\mathbf{o}$ at the end of the tuple $(g, pc, \mathbf{m}, i, \mathbf{s}, \mathbf{o})$ to keep the same order as in formulas 141 to 146.

image

Here are the overall suggestions
```suggestion
The machine state $\boldsymbol{\mu}$ is defined as the tuple $(g, pc, \mathbf{m}, i, \mathbf{s}, \mathbf{o})$ which are the gas available, the program counter $pc \in \mathbb{N}_{256}$ , the memory contents, the active number of words in memory (counting continuously from position 0), the stack contents, and the returned data buffer. The memory contents $\boldsymbol{\mu}_{\mathbf{m}}$ are a series of zeroes of size $2^{256}$.
```

Paper.tex
```diff
@@ -2364,8 +2364,9 @@ \subsection{Instruction Set}
 &&&& $\boldsymbol{\mu}'_{\mathbf{s}}[0] \equiv \lVert \boldsymbol{\mu}_{\mathbf{o}} \rVert$ \\
 \midrule
 0x3e & {\small RETURNDATACOPY} & 3 & 0 & Copy output data from the previous call to memory. \\
+&&&& reverts if $\quad \boldsymbol{\mu}_{\mathbf{s}}[1] + \boldsymbol{\mu}_{\mathbf{s}}[2] > \lVert\boldsymbol{\mu}_{\mathbf{o}}\rVert$\\
 &&&& $\forall i \in \{ 0 \dots \boldsymbol{\mu}_{\mathbf{s}}[2] - 1\}: \boldsymbol{\mu}'_{\mathbf{m}}[\boldsymbol{\mu}_{\mathbf{s}}[0] + i ] \equiv
-\begin{cases} \boldsymbol{\mu}_{\mathbf{o}}[\boldsymbol{\mu}_{\mathbf{s}}[1] + i] & \text{if} \quad \boldsymbol{\mu}_{\mathbf{s}}[1] + i < \lVert \boldsymbol{\mu}_{\mathbf{o}} \rVert \\ 0 & \text{otherwise} \end{cases}$\\
+\boldsymbol{\mu}_{\mathbf{o}}[\boldsymbol{\mu}_{\mathbf{s}}[1] + i]$\\
```
pldespaigne, line 2369: I would suggest to revert those changes.
The `RETURNDATACOPY` behavior is correct and don't need to be fixed.
The *Exceptional Halting* function (9.4.2.) already handle the case where we try to read past the returned data buffer.

image

pldespaigne (changes requested):
Good catch!
There is no reason for the returned data buffer $\mathbf{\mu}_\mathbf{o}$ to not be included in the *Machine State* section (9.4.1.).

However the `RETURNDATACOPY` specs are correct and don't need to be modified imo.

Paper.tex
RenanSouza2: I would still argue for altering the retiurndatacopy because it is a bit odd to include a case that will not happen,

what do you think about removing this line and keep the other alterations
image

Paper.tex
```diff
@@ -1187,7 +1187,7 @@ \subsection{Execution Overview}
...
 
 \subsubsection{Machine State}
-The machine state $\boldsymbol{\mu}$ is defined as the tuple $(g, pc, \mathbf{m}, i, \mathbf{s})$ which are the gas available, the program counter $pc \in \mathbb{N}_{256}$ , the memory contents, the active number of words in memory (counting continuously from position 0), and the stack contents. The memory contents $\boldsymbol{\mu}_{\mathbf{m}}$ are a series of zeroes of size $2^{256}$.
+The machine state $\boldsymbol{\mu}$ is defined as the tuple $(g, pc, \mathbf{m}, i, \mathbf{s}, \mathbf{o})$ which are the gas available, the program counter $pc \in \mathbb{N}_{256}$ , the memory contents, the active number of words in memory (counting continuously from position 0), the stack contents, and the returndata cache. The memory contents $\boldsymbol{\mu}_{\mathbf{m}}$ are a series of zeroes of size $2^{256}$.
```
pldespaigne, line 1190: You still forgot to rename "cache" into "buffer"

Paper.tex
pldespaigne: The case where we read outside of the returned data buffer should never happen (because $Z$ is executed beforehand).
It was only described for completeness and that's why we should leave it as is imo. It also guarantee this property (exception should be caught beforehand)

image

0xConwayGlider (approved):
SGTM, nice work!

yperbasis (approved):

Issue #910, opened 2024-07-17
The accrued state after an unexceptional but aborted `to_address` of a CALL


OlivierBBB:
- [Description](#description)
   * [In the Yellow paper](#in-the-yellow-paper)
   * [In EIP-2929](#in-eip-2929)
   * [In the execution specs](#in-the-execution-specs)
   * [In Besu](#in-besu)
- [Solution](#solution)

<a name="description"></a>
# Description

This issue is a follow-up on [an issue documented on the execution-specs repo](https://github.com/ethereum/execution-specs/issues/978). There seems to be a discrepancy between the yellow paper and the aforementioned implementation when it comes to updating the accrued state $A$ when a CALL-type instruction
- raises no exception
- but is aborted in the sense that either
  - the instruction attempts to transfer more value than the current balance allows for or
  - the current frame has call stack depth  1024

<a name="in-the-yellow-paper"></a>
## In the Yellow paper

The latest version of the Yellow Paper (Yellow paper `PARIS VERSION 71beac3`), as well as all versions stretching back at least to London, sees the accrued state unmodifed in the case described above i.e. in the _otherwise_ case depicted below

<img width="1535" alt="image" src="https://github.com/user-attachments/assets/9f54cfcd-6147-4849-840b-ca32e9577b8b">

In the "_otherwise_ case" the new accrued substate $A'$ would be $A$, the accrued substate prior to processing the CALL-type instruction, is just the original accrued substate, i.e. $A' = A$.

<a name="in-eip-2929"></a>
## In EIP-2929

Yet [EIP-2929](https://eips.ethereum.org/EIPS/eip-2929) suggests that one should have $A' = A ^ *$. See the following paragraph

<img width="1049" alt="image" src="https://github.com/user-attachments/assets/c6d53d69-7a7e-454f-897b-6ecda417b536">

This is what I understand to be implemeted in the execution specs (see the two links below) and in Besu as well (see link below.) Also this interpretation (warm the address regardless of whether or not the EVM 'enters' the CALL) makes intuitively more sense. Indeed the CALL pricing takes into account the cost of warming the recipient. Since we are considering the unexceptional case there will be no out of gas exception and the full price of the CALL instruction (modulo the possible $G_\text{callstipend}$ immediate gas refund) will be deducted from the available gas and execution will continue in the current context. It only makes sense that the address which one paid the warming of should remain warm after the aborted CALL-type instruction.

<a name="in-the-execution-specs"></a>
## In the execution specs

[...]

PR #915, opened 2024-08-25
add missing A* to CALL-type opcodes

pldespaigne:
In PR #913, fixing issue #910, $A^*$ was added if call is aborted.

image

However that PR forgot to add the same $A^*$ to other call type opcodes such as DELEGATECALL & CALLCODE.
This PR fix it.

Note: We don't need to make the same update to STATICCALL, as it is defined using CALL specs.


yperbasis (approved):
Good catch, thank you!

PR #917, opened 2024-08-29
style: add some missing words.

blablalf:
Hello, while reading the paper, I noticed a few small typos. Here is a fix if you want it.

- The phrase "*must at least as high as*" should be corrected to "*must **be** at least as high as.*".
- The phrase "*Gtxcreate is added if the transaction is contractcreating, but not if a result of EVM-code.*" should be "*Gtxcreate is added if the transaction is contractcreating, but not if **it is** a result of EVM-code.*".


yperbasis (changes requested):
The original text is confusing/incorrect. It should say something like "but not if it is a message call" instead.

blablalf:
> but not if it is a message call

I thought the sentence you recommended was clear, so I simply commited it as it was.

yperbasis (approved):

Issue #878, opened 2022-10-15
Maintain previous versions of YP under some directory in this repo


behkod:
Currently only the latest version of the YellowPaper (built based on Master branch) is available for easy download. In order to have access to prev. versions (according to `BRANCES.md` file) one needs to clone the repo, install additional LaTeX tools (~2GB of data) & work through to make their target version available.

I believe simply maintaining a pdf version of previous versions under this repo (maybe in some `dist` category or the same `build` dir) could make the process much much easier at a very low cost. Some researchers may need to refer to prev. versions for research/historical reasons.

chanhosuh:
Git tags would be very easy to do and accomplishes basically the same thing.

chfast:
> Git tags would be very easy to do and accomplishes basically the same thing.

Yes, this make sense. You can use GitHub releases to store the generated PDFs.

u59149403:
Please, maintain all versions for hardforks in same branch.

Execution client still need to verify old blocks. Thus old versions of YP still needed for clients. They don't have merely historic value. They are needed for implementing clients!

And every time we, say, fix some typo, this change need to be applied to all hardforks.

So, please make all versions available in same branch.

I would create PR myself, but I cannot do this, because:
- Such PR also need to change process of generation of artifacts for ethereum.org site. Ideally we should generate pdfs for all old versions and place them on ethereum.org . But I don't know how CI works, so I cannot make such change
- We need not only to store everything in `master`, but also to delete old branches. I have no rights to do this

u59149403:
This will enable us to fix errors in old versions in one single PR.

**Also: execution specs do the same!** They store everything in single branch and often change all them. For example, this is recent commit to execution specs, which fixes all versions at once: https://github.com/ethereum/execution-specs/commit/6b0f19f72cbc8458620ad313b6863ed0ea941412

PR #927, opened 2025-06-04
There is no any sense in having G_zero, which is equal to 0. This is well-known programming anti-pattern

u59149403:
There is no any sense in having G_zero, which is equal to 0. This is well-known programming anti-pattern


u59149403:
Of course, this change should be applied to all prev. hardforks, but it is difficult to do so because of https://github.com/ethereum/yellowpaper/issues/878

